PR #10, opened 2017-10-09
Fix url escapes

rickh94:
Table names containing slashes cause an invalid url to be created and any get request throws an error. Using urllib.parse.quote to escape the name fixes this issue.

I've included a test for my code but unfortunately I couldn't run the main test suite. I used the library with this patch applied to download some records and it seemed to work fine, so it doesn't seem like I broke anything.

### DIFF
--- a/airtable/airtable.py
+++ b/airtable/airtable.py
@@ -22,6 +22,7 @@ from requests.exceptions import HTTPError
 import posixpath
 import time
 from six.moves.urllib.parse import unquote
+from six.moves.urllib.parse import quote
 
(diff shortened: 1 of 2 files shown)


gtalarico:
Thank you @rickh94 much appreciated.
And my apologies about the test suite. I need to convert all tests to mocked requests but haven't had the time.

rickh94:
My pleasure. 

If I have time I'd be happy to write some tests. Your library provides some really nice functionality that saved me quite a bit of time.

PR #17, opened 2018-02-08
Added typecast argument to insert()

MADindustries:
This allows Airtable to perform typecasting if necessary to fit data into a record. This is disabled by default.


gtalarico:
Merged, thank you !

PR #38, opened 2019-06-05
Issue3 - Full Test Suite using Mocks

bpeterso2000:
A full suite of unit tests that don't require API keys, Airtable setup, connectivity, etc.  100% code coverage + tox for py27, py35, py36 & py37.


gtalarico:
Hey - thank you so much for submitting this.

Unfortunately, I did a bunch of duplicate work on this week, unaware you were working on the same things.

I would still be happy to accept this, but we would need to reconcialte with the work I already merged.

In other words, can we try to resolve conflicts with Master?
Can you try rebase from master and see how big the conflict is?

Some of the issues I noticed:

* Tox was also added to master - were both approached the same?

* I also added parametrized Url Params tests, however we are using different styles - you check the param dict while I check the actual encoded url. I think you approach might be better, since we should not be testing the url encoding, but only the params formatting.

* Mock tests - I started using `requests_mock` instead of `Mock.` I would be ok with using Mock as well, although requests_mock seems easier to read to me.

* `validate_session` on class init was removed. This was unecessary and should make testing easier (no mock for tests that don't fire requests)

Thanks again, let's work through these to get this merged

Issue #42, opened 2019-06-08
Formular Filter fails with boolean value


gtalarico:
Given [this base](https://airtable.com/shrREAyVIxHiBEZOF), I should be able to:
```
records = airtable.search('On Display?', True)
```
Instead I get 
```
HTTPError: 422 Client Error: Unprocessable Entity for url: https://api.airtable.com/v0/appNPoAJOyctb4MId/Artists?filterByFormula=%7BOn+Display%3F%7D%3DTrue
```

While you are at it, `records = airtable.search('On Display?', "1")` returns all records, should return `[]`

gtalarico:
New version #126 casts bool values to int to fix this.

PR #43, opened 2019-06-10
Add an example of how to search by checkbox

maxko87:
A common use-case of `search` is to filter all fields where a checkbox is checked/unchecked. Counterintuitively, passing any of "True" (string) or True (boolean) doesn't work. The way to get around this is by passing 1 or 0. This is undocumented until now.


gtalarico:
Thanks! Can you correct the linting error?

Related to #42

gtalarico:
I know the boolean field records actually return boolean values, but the search methods is using the formula which might expect 0 or 1 values. 

Maybe we could handle this when building the formula parameter? 

https://support.airtable.com/hc/en-us/articles/203255215-Formula-field-reference

Issue #39, opened 2019-06-06
do not limit an airtable instance to a single table_name


bendidi:
if we have multiple tables in a `base_key`, it's kind of frustrating to have to create a different `Airtable` instance for each table, what I propose is to make `table_name` a variable in each `insert`, `update`, `delete` call, the changes are quite minim: 
```python
def __init__(self, base_key, api_key=None):
        """
        If api_key is not provided, :any:`AirtableAuth` will attempt
        to use ``os.environ['AIRTABLE_API_KEY']``
        """
        session = requests.Session()
        session.auth = AirtableAuth(api_key=api_key)
        self.session = session
        self.base_url = posixpath.join(self.API_URL, base_key)
```

```python
def record_url(self, table_name, record_id):
        """ Builds URL with record id """
        url_safe_table_name = quote(table_name, safe="")
        return posixpath.join(self.base_url, url_safe_table_name, record_id)
```

and a method could be updated as so :
```python
def get(self, table_name, record_id):
        """
        Retrieves a record by its id
        >>> record = airtable.get('recwPQIfs4wKPyc9D')
        Args:
            record_id(``str``): Airtable record id
        Returns:
            record (``dict``): Record
        """
        record_url = self.record_url(table_name, record_id)
        return self._get(record_url)
```

if you are okay with the changes I can make a PR

gtalarico:
Thanks for submitting this. 
I have an open issue to make table name settable, so I am glad you opened this.
See #13 

I think it's fair use case however, I also think it would be equally frustrating if you only use one table, and you have to enter the table name on each request, don't you agree?

Let's consider some options:
* Each class instance is associated with a table - current state, good for single table use
* Class instance is not linked to a single table, and a table name is provided on each request - your proposal, good for multiple tables use
* Instance _can_ take a default table name (kwargs), but requests can also pass an _optional_ table name  - most flexible, but seems it could get confusing.
* Subclass Airtable to a create a Class that takes table name on each request. This seems the most flexible, and most importantly, won't break backward compatibility. 
Not sure how we would name the subclass.

For the time being, you can also just create this subclass yourself with the changes you proposed.

Thoughts?

bendidi:
> I think it's fair use case however, I also think it would be equally frustrating if you only use one table, and you have to enter the table name on each request, don't you agree?

[...]

PR #47, opened 2019-06-17
create a AirtableBase for a general airtable wrapper that can accept `table_name` as argument

bendidi:
as discussed in #39 

this PR :

- adds a base class for creating an `AirtableBase` instance independent of `table_name`
- keep the old `Airtable` class for backward compatibility
- refactor the two classes
- make sure all tests pass

### DIFF
--- a/airtable/airtable.py
+++ b/airtable/airtable.py
@@ -92,6 +92,7 @@ similar to this:
 
 import sys
 import requests
+from functools import partial
 from collections import OrderedDict
 import posixpath
 import time
@@ -107,14 +108,14 @@ except AttributeError:
     IS_IPY = False
 
 
-class Airtable:
+class AirtableBase:
 
     VERSION = "v0"
     API_BASE_URL = "https://api.airtable.com/"
     API_LIMIT = 1.0 / 5  # 5 per second
     API_URL = posixpath.join(API_BASE_URL, VERSION)
 
-    def __init__(self, base_key, table_name, api_key=None):
+    def __init__(self, base_key, api_key=None):
         """
         If api_key is not provided, :any:`AirtableAuth` will attempt
         to use ``os.environ['AIRTABLE_API_KEY']``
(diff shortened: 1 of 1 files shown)


airtable/airtable.py
```diff
@@ -1,582 +1,918 @@
...
+            sort (``list``, optional): List of fields to sort by.
+                Default order is ascending. See :any:`SortParam`.
+            formula (``str``, optional): Airtable formula.
+                See :any:`FormulaParam`.
+
+        Returns:
+            iterator (``list``): List of Records, grouped by pageSize
+
+        """
+        return self.get_iter_in_table(self.table_name, **options)
```
gtalarico: I am not convinced we should modify the method names.

Airtable methods could just super()` call `Airtable` method and pass table name

airtable/airtable.py
```diff
@@ -1,582 +1,918 @@
...
+
+    def __init__(self, base_key, table_name, api_key=None):
+        """
+        If api_key is not provided, :any:`AirtableAuth` will attempt
+        to use ``os.environ['AIRTABLE_API_KEY']``
+        """
+        super().__init__(base_key, api_key=api_key)
+        self.table_name = table_name
+
+        # the 2 lines below are not really needed but kept just for passing tests
```
gtalarico: Can we fix tests instead?

gtalarico:
Why is the diff so large? First part has only minor changes, we should be seeing only changes highlighted:
image

bendidi:
no idea why the diff is so large, the only thing I did was add the new class `AirtableBase`, and modify the old class `Airtable`

airtable/airtable.py
bendidi: the method name has not been modified in `Airtable`, but just in `AirtableBase`, and it's kind of a refactor to not repeat the same code twice

bendidi: `Airtable` still have the same method name, and I think that for `AirtableBase` adding the `in_table` to each method is better, to make the user aware that it's a method that queries from a table that he should specify

airtable/airtable.py
bendidi: yes tests can be easily fixed, but it's more for backward compatibility 
more info [here](https://github.com/gtalarico/airtable-python-wrapper/pull/47/files/8df390a1c2054f0dc35385c5ed71f57953a8adc9#diff-6de8259f9857d7a7fa0a29f9cbb124dcR554)

airtable/airtable.py
gtalarico: I find this pattern confusing, especially because subclass would also have all of the parents class. I would prefer to use method inheritance. It's ok if signatures are different. 
See example.

What do you think?

````
AirtableBase():
    def get(self, table_name, record_id):
        record_url = self.record_url(record_id)
        return self._get(record_url)

Airtable():
    def get(self, record_id):
        return super()._get(self.table_name, record_id)

table = Airtable(x, y, z )
table.get(record_id)

base = AirtableBase(x, y)
base.get(table_name, record_id)
```

bendidi: agree, yes it can be less confusing

yun-wang:
The large diff is most likely caused by https://help.github.com/en/articles/configuring-git-to-handle-line-endings.

gtalarico:
Created issue to track this #71

PR #57, opened 2019-08-21
setup.* cleanup, compatibility fixes

dargueta:
### Significant changes

* Drop support for Python 3.4 -- it's no longer maintained and we weren't testing it anyway
* Add official support for Python 3.7 and 3.8
* Block installation on unsupported Python versions using `python_requires` directive

### Testing

* Make tox environments consistent with what we claim to support
* Use tox in Travis
* Add `.coverage` artifact to gitignore

### Compatibility

* Don't use old-style classes on Python 2
* Python 2 doesn't support `encoding` kwarg in `open()`
* Python 2 doesn't have `json.JSONDecodeError`; use its superclass instead

### General cleanup

* Remove most I/O in setup.py by moving description to setup.cfg
* Consolidate installation requirements
* Run tests through `black` like we do with the rest of the codebase


gtalarico:
This is great. Thank you and sorry for the delay in merging it

gtalarico:
Thank you for your help @dargueta 
It's been released on https://github.com/gtalarico/airtable-python-wrapper/releases/tag/0.13.0

PR #65, opened 2020-01-03
Update airtable.py

johruss:
Hello, as said in the request.request, a timeout should be specified.
I have an unstable internet connection, and this connector will hand for an unspecified amount of time.
Putting the timeout parameter solves this problem.

### DIFF
--- a/airtable/airtable.py
+++ b/airtable/airtable.py
@@ -166,7 +166,7 @@ class Airtable:
         """ Builds URL with record id """
         return posixpath.join(self.url_table, record_id)
 
-    def _request(self, method, url, params=None, json_data=None):
+    def _request(self, method, url, params=None, json_data=None, timeout=(10, 30)):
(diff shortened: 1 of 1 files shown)


gtalarico:
@johruss don't you need to include the param in the call to requests session?

johruss:
> @johruss don't you need to include the param in the call to requests session?

I had a blonde moment :)
Sorry. I've corrected that.

gtalarico:
Actually, not sure hard coding different defaults for timeouts here is the best way to go about this.
Maybe we can store a `Airtable._timeout` tuple that can be configured by the user during init.


```
table = Airtable(..., timeout=(10,30))
```


ps: that's how pyGithub did it
https://pygithub.readthedocs.io/en/latest/github.html#main-class-github

johruss:
Yes, this may be better.
Is this something you could do?

PR #77, opened 2020-03-26
Rewrites delete() and batch_delete() to support bulk delete per request

joegnis:
Hi, I came upon this repo and found it super useful in my personal project. The new bulk support of Airtable API looks nice, so I decided to give it a try to add it to this library. 

I tried not to change the behavior of existing public methods. I found that batch_xxx methods call `_batch_request()`, so to not change too much code, I decided to let `delete()` accepts multiple ids, and made `batch_delete()` call it repeatedly. Since `batch_delete()`'s behavior changed so much, I had to change the test for it.

I also fixed some typos in the tests along the way. If this PR goes well, I would like to add bulk support to methods as well.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,8 +10,11 @@ docs/build
 .tox
 *egg-info
 *__pycache__
+*.pyc
 .pytest_cache
 .cache
 coverage.xml
 .coverage
 test.py
+.python-version
+.mypy_cache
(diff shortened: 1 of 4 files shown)


gtalarico:
I think batch api has been added in some form.
sorry we weren't able to use this, but thank you for your contribution.

PR #78, opened 2020-03-30
Adds support to cellFormat, timeZone and userLocale as per documentation

joao-marques-at-databeyond:
Hi! 
I needed cellFormat for a "copy across bases" functionality, so I coded it into your project. In case you want to integrate it :)
Cheers, and thank you for your library!
Joao

### DIFF
--- a/airtable/params.py
+++ b/airtable/params.py
@@ -160,6 +160,99 @@ class AirtableParams(object):
         param_name = "pageSize"
         kwarg = "page_size"
 
+    class CellFormatParam(_BaseParam):
+        """
+        cellFormat Param
+
+        Kwargs:
+            ``cellFormat=``
+
+        If set, the format that should be used for cell values. Supported values are:
+           json: cells will be formatted as JSON, depending on the field type.
+           string: cells will be formatted as user-facing strings, regardless of the field type.
+                   Note: You should not rely on the format of these strings, as it is subject to change.
+
+        The default is json.
+
+        Usage:
+
+        >>> airtable.get_all(cellFormat='json')
+
+        Args:
+            cellFormat (``str``): The format to use.
+
+        """
+
+        # Class Input > Output
+        # >>> filter = CellFormatParam('The format to use.')
+        # >>> filter.to_param_dict()
+        # {'cellFormat: 'The format to use.'}
+
+        param_name = "cellFormat"
+        kwarg = param_name
+
+    class TimeZoneParam(_BaseParam):
+        """
+        timeZone Param
+
+        Kwargs:
+            ``timeZone=``
+
+        The time zone that should be used to format dates when using string as the cellFormat.
+        This parameter is required when using string as the cellFormat.
+        List of accepted values:
+        https://support.airtable.com/hc/en-us/articles/216141558-Supported-timezones-for-SET-TIMEZONE
+
+        No default. It is MANDATORY WHEN USING string as CellFormat.
+
+        Usage:
+
+        >>> airtable.get_all(timeZone='Europe/London')
+
+        Args:
+            timeZone (``str``): One of https://support.airtable.com/hc/en-us/articles/216141558-Supported-timezones-for-SET-TIMEZONE
+
+        """
+
+        # Class Input > Output
+        # >>> filter = TimeZoneParam('The timeZone to use.')
+        # >>> filter.to_param_dict()
+        # {'timeZone: 'The timeZone to use.'}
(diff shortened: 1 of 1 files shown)


gtalarico:
Thanks. 
Can we check line ending and make sure it uses same as in editorconfig file? 
Should be LF. Hopefully that's what it's on the source code. 

Diff as is make it hard to actually view pr

joao-marques-at-databeyond:
Hi, sorry it's super annoying, but I'm using windows and using gitextensions which seems to make a bit of a witchcraft on it. You can see differences only if you go to the config wheel next on the top left next to "Jump to", and select "Hide Whitespace Changes". Or if you want, I am happy for you to copy the code and create a pull in your name. 
Thanks! Cheers

airtable/params.py
```diff
@@ -1,368 +1,462 @@
...
+        kwarg = "page_size"
+
+    class CellFormatParam(_BaseParam):
+        """
+        cellFormat Param
+
+        Kwargs:
+            ``cellFormat=``
+
+        If set, the format that should be used for cell values. Supported values are:
```
gtalarico, line 170: seems like supported value are only `string` and `json`, correct?
maybe we should check/validate input to allowed values?

airtable/params.py
```diff
@@ -1,368 +1,462 @@
...
+
+        """
+
+        # Class Input > Output
+        # >>> filter = CellFormatParam('The format to use.')
+        # >>> filter.to_param_dict()
+        # {'cellFormat: 'The format to use.'}
+
+        param_name = "cellFormat"
+        kwarg = param_name
```
gtalarico, line 192: `kwarg` can be used to allow for standard python formatting eg (snake case) or more concise param. name. For example, here we could use `kwarg = cell_format` or event `kwarg = format`
These were implemented so user can always use the official airtable param names, but can also use custom library "aliases" for cleaner python-styled kwarg names

eg. `param_name = filterByFormula=` -> `kwarg= formula`

airtable/params.py
```diff
@@ -1,368 +1,462 @@
...
+
+    class CellFormatParam(_BaseParam):
+        """
+        cellFormat Param
+
+        Kwargs:
+            ``cellFormat=``
+
+        If set, the format that should be used for cell values. Supported values are:
+           json: cells will be formatted as JSON, depending on the field type.
```
gtalarico, line 171: I read the docs but I don't actually understand what this configures.
do you have an example for payloads with and without this option?

airtable/params.py
```diff
@@ -1,368 +1,462 @@
...
+
+        """
+
+        # Class Input > Output
+        # >>> filter = TimeZoneParam('The timeZone to use.')
+        # >>> filter.to_param_dict()
+        # {'timeZone: 'The timeZone to use.'}
+
+        param_name = "timeZone"
+        kwarg = param_name
```
gtalarico, line 223: similar to comment above - `kwarg = timezone`?

airtable/params.py
```diff
@@ -1,368 +1,462 @@
...
+
+        Kwargs:
+            ``userLocale=``
+
+        The user locale that should be used to format dates when using string as the cellFormat.
+        This parameter is required when using string as the cellFormat.
+        List of accepted values:
+        https://support.airtable.com/hc/en-us/articles/220340268-Supported-locale-modifiers-for-SET-LOCALE
+
+        No default. It is MANDATORY WHEN USING string as CellFormat.
```
gtalarico, line 237: does the api have clear msg when these are missing?
If so, maybe we should validate?

gtalarico:
> Hide Whitespace Changes
Did not know this existed! Thanks for showing me that - I was able to review the code.

In any case, we need to fix line endings before merging, otherwise we will keep getting PRs with line ending changes. 

What editor are you using? Most editors have plugins set line endings based on `.editorconfig` file.
As for git, I had this issue for while as a former windows user. I now set git config to `core.autocrlf=input` always. This leaves files untouched by git, and it's up to editors to decided line endings - I like this being a more explicit option than the other auto-convert settings.

Anyhow, thank you for this PR. I would like to get this merge, pending issues commented

craigmonson:
You can force a project to normalize line endings: https://docs.github.com/en/github/using-git/configuring-git-to-handle-line-endings#per-repository-settings

gtalarico:
Created issue to cover this. 
closing PR because of 1.x release this would need a total re-write
Thanks

Issue #94, opened 2020-08-19
422 error when batch deleting 1 record


cuchoi:
Hi! Thanks for this amazing wrapper.

I think the recent update to support the limit of ten records ids broke the ability to batch delete one record. Is this expected? 
Before I used to do:
```
record_ids = [record["id"] for record in records]
airtable.batch_delete(record_ids)
```
But I started getting the following error:
```
requests.exceptions.HTTPError: ('422 Client Error: Unprocessable Entity for url: https://api.airtable.com/v0/<app_id>/Reporting?records=recXXXXXXXXXXX', '422 Client Error: Unprocessable Entity for url: https://api.airtable.com/v0/<app_id>/Reporting?records=recXXXXXXXXXXX [Error: {\'type\': \'INVALID_RECORDS\', \'message\': \'"records" must be an array of up to 10 record IDs.\'}]')
```
Now I am doing
```
record_ids = [record["id"] for record in records]
if len(record_ids) == 1:
      connection_airtable.delete(record_ids[0])
else:
      connection_airtable.batch_delete(record_ids)
```
Not a big deal but I thought it might be useful to report this issue.

KNWR:
Hi, thanks for creating the library. Wanted to confirm I'm finding the same error.

gtalarico:
thank you for confirming this. I will try to take a look at this when I can, in the meantime if someone has the time to investigate this and prepare a PR fix I am happy to review and merge it

KNWR:
cc @gtalarico we can probably close this?

gtalarico:
yep thanks!

PR #100, opened 2020-09-24
Fix 422 error when calling batch_delete on a single record

KNWR:
Surfaced in issue #94 . (cc @cuchoi)

The Airtable API has two deletion endpoints. The batch endpoint deletes multiple records from a table at once, while the record endpoint deletes one record. 

The API documentation says to "issue a DELETE request to the record endpoint to delete a single record." The batch_delete function errors because it sends DELETE requests with one record to the batch endpoint. 

The batch_delete function takes multiple records and sends them to the batch endpoint in chunks of up to n records at a time. The error case occurs if a chunk has just one record. This happens in two cases: i) the method is called on just one record, or ii) number_of_records > n with number_of_records % n == 1, i.e. after splitting the records into chunks of size n there is one record remaining, leaving a final chunk with 1 record.

These changes resolve the error by checking if a chunk contains just one record and, if so, calling the record deletion endpoint instead of the batch endpoint. They also add test coverage for calling batch_delete on a single record.

### DIFF
--- a/airtable/airtable.py
+++ b/airtable/airtable.py
@@ -203,6 +203,9 @@ class Airtable(object):
         return self._request("delete", url)
 
     def _delete_batch(self, record_ids):
+        if len(record_ids) == 1:
+            return self.delete(record_ids[0])
+
         return self._request("delete", self.url_table, params={"records": record_ids})
 
     def get(self, record_id):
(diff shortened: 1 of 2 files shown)


gtalarico:
This looks good to me, thank you so much for taking this on!

gtalarico:
I will test this tonight locally and push a new release

KNWR:
🍻!

gtalarico:
Released here
https://pypi.org/project/airtable-python-wrapper/0.15.1/

PR #107, opened 2020-10-23
Batch update

thomasopsomer:
As the title says, just a simple method to update records in batch :)

And line ending should be ok this time :)

### DIFF
--- a/airtable/airtable.py
+++ b/airtable/airtable.py
@@ -435,6 +435,25 @@ class Airtable(object):
             record_url, json_data={"fields": fields, "typecast": typecast}
         )
 
+    def batch_updates(self, records, typecast=False):
+        """
+        Updates a records by their record id's in batch.
+
+        Args:
+            records(``dict``): List of dict: {"id": record_id, "field": fields_to_updates_dict}
+
+        Returns:
+            records(``list``): list of updated records
+        """
+        updated_records = []
+        for chunk in self._chunk(records, self.MAX_RECORDS_PER_REQUEST):
+            response = self._patch(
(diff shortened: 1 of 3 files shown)


airtable/airtable.py
```diff
@@ -435,6 +435,26 @@ def update(self, record_id, fields, typecast=False):
             record_url, json_data={"fields": fields, "typecast": typecast}
         )
 
+    def batch_update(self, records, typecast=False):
+        """
+        Updates a records by their record id's in batch.
+
+        Args:
+            records(``dict``): List of dict: [{"id": record_id, "field": fields_to_update_dict}]
```
gtalarico, line 443: Could you add missing `Arg` documentation?
https://github.com/gtalarico/airtable-python-wrapper/blob/d829743342c34e11e1fb38fc5579a0533af4900a/airtable/airtable.py#L524

gtalarico (changes requested):
Looks great, thank you for the PR!
Just need `Arg` change on docstring

thomasopsomer:
Sorry I forgot about this... just fixed the docstring

gtalarico (approved):
LG, thanks!
We should probably include "Default is `False" in doc string but we can fix this in sep. PR

DylanScotney:
Hey @gtalarico do you plan on doing a new release to include this? Would be great to have it as part of the pypi library!

gtalarico:
Thanks for the bump @DylanScotney 
released as .15.12
https://pypi.org/project/airtable-python-wrapper/#history

PR #111, opened 2021-03-09
Escape single quotes in formula field_value

kevindalias:
I encountered a bug where Airtable responds with a 422 error when a formula value contains a single quote or apostrophe. For example, a string like "Men's Clothing" is incompatible with the single quotes used to wrap the formula. Escaping the single quote resolved the issue. This library makes my life better on a daily basis so the least I can do is contribute a (very small) improvement. Hope this is useful, thanks!

### DIFF
--- a/airtable/params.py
+++ b/airtable/params.py
@@ -205,7 +205,7 @@ class AirtableParams(object):
             Creates a formula to match cells from from field_name and value
             """
             if isinstance(field_value, str):
-                field_value = "'{}'".format(field_value)
+                field_value = "'{}'".format(field_value.replace("'", "\\'"))
 
             formula = "{{{name}}}={value}".format(name=field_name, value=field_value)
             return formula


gtalarico:
Thank you for submitting this

1. We should have some tests for these escaping cases
2. Wonder if this could break someone's code who is already escaping due to previous behavior... 
Maybe we should replace only when is not being escaped?

What do you think

kevindalias:
Those are great points, thank you. I think you're right on both of them.

1. Yes, we should have tests for these cases. I will take a look at the tests you already have and try to extend them/add a test for the escaping case. I haven't actually worked with pytest before, so it'll be great to learn it. I'll take a pass after addressing your second point.
2. Yes, you're right. It's definitely possible people have already worked around this issue. I think regex might be the easiest way to detect it, but it would add an `re` import. Do you think it'd be better to avoid another import and try for something simpler? Let me know which approach you think would be better.

gtalarico:
> I haven't actually worked with pytest before
awesome, let me know if you have any issues with dev setup. I thought I had a `CONTRIBUTING.md` but looks like it's missing. `pip install -e .` + `pip install -r requirements-dev.txt` should work, or `python setup.py development`, don't recall if that one install dev pkgs like pytest 

> would add an re import.
no problem using builtin `re` !

kevindalias:
Sounds good to me! Thanks for the tips on setup. I'm also glad to hear using builtin `re` works for you. I'll give this a go and will follow up when ready (or if I have any blocking questions).

kevindalias:
Closing because I submitted a new pull request from my fork. https://github.com/gtalarico/airtable-python-wrapper/pull/120

PR #114, opened 2021-04-20
move base_key to base_id to match airtable nomenclature

rjmoggach:
I didn't update the notebook but the library and docs now match the airtable api nomenclature.

I would suggest moving the library to `python-airtable` to create a pypi semantically pretty package name.


gtalarico:
Thanks @rjmoggach 

>I would suggest moving the library to python-airtable to create a pypi semantically pretty package name.

Would love to have picked a different name but I am not sure it's worth changing it now (work involved vs value)
and given I am the only maintainer here :)

Also looks like a fork is currently using your [suggested name](https://pypi.org/project/python-airtable/)

PR #115, opened 2021-04-29
Fix typo in documentation for batch_update

Richie78321:
Fixed a minor typo in the `batch_update` function documentation. Using `field` in the record dictionary leads to a key error that is solved by using `fields` instead.


Richie78321:
Not sure why the build failed on a 1-char change haha

gtalarico:
Thank you 
I think something else broke.
I will take a look at it

PR #117, opened 2021-06-11
Allow context manager

everettttt:
I'd like to use a `with` block:

```python3
from airtable import Airtable
from pandas import DataFrame

with Airtable("base_id", "table_name") as airtable:
    records = airtable.get_all()
df = DataFrame.from_records(r["fields"] for r in records)
```

I created minimal `__enter__()` and `__exit()__` methods. Let me know if you need anything else.

### DIFF
--- a/airtable/airtable.py
+++ b/airtable/airtable.py
@@ -138,6 +138,26 @@ class Airtable(object):
         self.url_table = posixpath.join(self.API_URL, base_key, url_safe_table_name)
         self.timeout = timeout
 
+    def __enter__(self):
+        """
+        Context manager setup logic. Called during `with` block setup.
+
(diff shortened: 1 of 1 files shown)


gtalarico:
Curious, what's the advantage of using a context manager here vs the code below?
AFAIX Context manager are using for managing actions that need to be taken before/after, eg. closing a file after opening
In this case the Airtable class has no persistent connection and nothing that needs to be done before/after making the call

Ie, this code should have the same effect:
```python
from airtable import Airtable
from pandas import DataFrame

airtable = Airtable("base_id", "table_name"):
records = airtable.get_all()
df = DataFrame.from_records(r["fields"] for r in records)
```

PR #136, opened 2021-09-10
Indicate type checking compatibility, according to PEP 561.

### LINKED ISSUE #135: Add support for type checking
Currently, using mypy gives an error that the library doesn't have type hints.

    error: Skipping analyzing "pyairtable": found module but no type hints or library stubs

It would be nice to have support for it, as the type annotations are already done.

fmorato:
Fixes #135.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,3 +1,7 @@
 import setuptools
 
-setuptools.setup()
+setuptools.setup(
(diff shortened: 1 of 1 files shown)


gtalarico:
Thanks 🙏

PR #139, opened 2021-12-17
Fix metadata api

nafeger:
So, I would definitely not suggest that you accept this PR. But I'm not totally sure the fix.

Basically what I'm seeing is that sometimes the api_key variable is a string that is the actual api_key, and sometimes the value is the actual pyairtable.Api class. Feels like something fishy is happening with these @property or setter annotations.

Sorry, it's not a solid fix, but it might help get closer.

### DIFF
--- a/pyairtable/api/abstract.py
+++ b/pyairtable/api/abstract.py
@@ -38,7 +38,10 @@ class ApiAbstract(metaclass=abc.ABCMeta):
         self._api_key = value
 
     def _update_api_key(self, api_key: str) -> None:
-        self.session.headers.update({"Authorization": "Bearer {}".format(api_key)})
+        if api_key in api_key:
+            self.session.headers.update({"Authorization": "Bearer {}".format(api_key.api_key)})
+        else:
+            self.session.headers.update({"Authorization": "Bearer {}".format(api_key)})
 
     @lru_cache()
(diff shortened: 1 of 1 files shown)


gtalarico:
Can you include a short reproducible example of the issue you are having?

nafeger:
Yeah sure: 
```

    base = Base(api_key=Api("KEY"), base_id="BASE_ID")
    schema = get_base_schema(base=base)
    for t in schema.items():
        print(t)
```

returns a 401:

`requests.exceptions.HTTPError: ('401 Client Error: Unauthorized for url: https://api.airtable.com/v0/meta/bases/<BASE_ID>/tables', "401 Client Error: Unauthorized for url: https://api.airtable.com/v0/meta/bases/<BASE_ID>/tables [Error: {'type': 'AUTHENTICATION_REQUIRED', 'message': 'Authentication required'}]")`


because when you debug the header for authentication token is: 

`Authorization: Bearer <pyairtable.Api>`

gtalarico:
should be :
```
base = Base(api_key="<your_api_key>", base_id="<base-id>")
```

nafeger:
argh!, that makes sense. It's like it's using what I passed in.

nafeger:
sorry about that the type spec clearly says string.

gtalarico:
Glad you got it working 👍

Issue #134, opened 2021-08-31
Add support to cellFormat, timeZone and userLocale


gtalarico:
See #78 
Needs re-write

gtalarico:
Fixed in #140

PR #140, opened 2022-02-12
Add support for cellFormat, timeZone, and userLocale API params 

marks:
Adds support for `cellFormat`, `timeZone`, and `userLocale` Airtable API params (as `cell_format`, `time_zone`, and `user_locale`)

The main benefit here is to be able to specify `cell_format='string'` (which requires defining a user locale and time zone) so the Airtable API returns all cell values as strings which can be helpful when wanting to save records to CSV without needing to traverse Airtable objects which vary based on the field type. `string` cell format also returns a comma-separated list of linked records' primary field values (instead of an array of record IDs)

Example usage:
```
from pyairtable import Table
api_key = 'keyXXX'
table = Table(api_key, 'appXXX', 'tblXXX')

recordsAsStrings = table.all(cell_format='string', user_locale='en_US', time_zone='America/Chicago', max_records=2)
recordsAsJson = table.all(cell_format='json', max_records=2)
```

(Related to #134 / #78)

### DIFF
--- a/docs/source/api.rst
+++ b/docs/source/api.rst
@@ -282,6 +282,15 @@ have a corresponding ``kwargs`` that can be used with fetching methods like :met
    * - ``fields``
      - ``fields``
      - |kwarg_fields|
+   * - ``cell_format``
+     - ``cellFormat``
+     - |kwarg_cell_format|
+   * - ``user_locale``
+     - ``userLocale``
+     - |kwarg_user_locale|
+   * - ``time_zone``
+     - ``timeZone``
+     - |kwarg_time_zone|
 
 
 Formulas
--- a/docs/source/substitutions.rst
+++ b/docs/source/substitutions.rst
@@ -48,3 +48,19 @@
 
 .. |kwarg_typecast| replace:: The Airtable API will perform best-effort
     automatic data conversion from string values. Default is False.
+
+.. |kwarg_cell_format| replace:: The cell format to request from the Airtable
+    API. Supported options are `json` (the default) and `string`. 
+    `json` will return cells as a JSON object. `string` will return
+    the cell as a string. `user_locale` and `time_zone` must be set when using
+    `string`.
+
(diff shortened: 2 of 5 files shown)


marks:
Hiya @gtalarico - first off: I'm a huge fan of pyairtable -- thank you :bow:

I realize this is likely not a complete PR yet but I wanted to start the discussion.

Do you have any advice on additional tests to write and where to check that user_locale and time_zone are set if cell_format is set to `string`? (Airtable will return an error if they are not, but it would be nice to have that happen client-side too):
```
requests.exceptions.HTTPError: ('422 Client Error: Unprocessable Entity for url: https://api.airtable.com/v0/appXXX/tblXXX?cellFormat=string', "422 Client Error: Unprocessable Entity for url: https://api.airtable.com/v0/appXXX/tblXXX?cellFormat=string [Error: {'type': 'INVALID_REQUEST', 'message': 'timeZone and userLocale must be specified when using the string cellFormat'}]")
```

gtalarico:
This looks great overall @marks .
I will take a closer look when I am from a trip, but should be a great addition.

gtalarico:
This looks great @marks - I am going to merge and then do a follow up PR for version bump.

> Airtable will return an error if they are not, but it would be nice to have that happen client-side too
I am a bit divided on this. What do we really get from catching this and raising? I think most likely we would just raise a similar error than what's currently returned by the API: `timeZone and userLocale must be specified when using the string cellFormat`

On the other hand, this means we have to maintain this - e.g. maybe the API could start accepting a default value, and we would need to update the code.

I think for now I don't see any major benefits in handling the missing required args, so I would suggest we just keep this as is.
Cheers

gtalarico:
[Release as 1.1.0](https://pypi.org/project/pyairtable/1.1.0/)

marks:
Woot - thanks @gtalarico ! 🎉 🚀

Issue #142, opened 2022-02-28
Add option to retry rate-limited requests


jskrzypek:
We tend to get a lot of failed airtable requests due to the rate limits, and are using [Tenacity](https://tenacity.readthedocs.io/en/latest/) to retry these requests automatically, which usually works well... I've created a wrapper function to handle this generically:

```py
from pyairtable import Table
from tenacity import retry, wait_fixed, stop_after_attempt

FOO_TABLE = Table(os.environ.get("AIRTABLE_API_KEY"), BASE_ID, TABLE_NAME)


@retry(wait=wait_fixed(3), stop=stop_after_attempt(3))
def call_airtable(tbl, *args, method="all", **kwargs):
    return getattr(tbl, method)(*args, **kwargs)


def get_foo(foo_id):
    return call_airtable(FOO_TABLE, foo_id, method="get")
```

While this strategy generally works for us, it wraps the whole pyAirtable function and there are some functions in the pyAirtable API – i.e. [`iterate()`](https://pyairtable.readthedocs.io/en/latest/api.html#pyairtable.api.Table.iterate), [`all()`](https://pyairtable.readthedocs.io/en/latest/api.html#pyairtable.api.Table.all), [`batch_create()`](https://pyairtable.readthedocs.io/en/latest/api.html#pyairtable.api.Table.batch_create), & [`batch_update()`](https://pyairtable.readthedocs.io/en/latest/api.html#pyairtable.api.Table.batch_update) – where this strategy is suboptimal. It seems to me that it would be much better to just patch the retry functionality into the inner [`_request()`](https://github.com/gtalarico/pyairtable/blob/bb13d4c06e3917682c1900afbb14f88ba65adc5a/pyairtable/api/abstract.py#L91-L95) method of the ApiAbstract super class, so only the actual chunk or offset request that was rate-limited gets retried.

@gtalarico, would you be open to a PR implementing retry functionality on the inner API calls? I see you previously looked into this and closed the PR in #93. 

For comparison, [the official `airtable.js` JS library has this functionality built in](https://github.com/Airtable/airtable.js/blob/9d40666979af77de9546d43177cab086a03028bf/src/run_action.ts#L70-L75).

[...]

PR #143, opened 2022-03-01
Introduce RequestStrategy pattern

jskrzypek:
[Implements #142]

This PR introduces a new pattern for consumers to inject alternate request
functionality into pyAirtable's ApiAbstract base class. A RequestStrategy should
accept the request parameters pyAirtable uses, make the HTTP request to the
Airtable API and then process the response and return the contained records in
the format that pyAirtable expects.

Three implementations are included, SimpleRequestStrategy,
RetryingRequestStrategy, and RateLimitRetryingRequestStrategy:

SimpleRequestStrategy encapsulates the use of a `requests.Session` to make the
requests, and expects the caller to provide any needed parameters like timeout
or headers. This behavior should be a 1:1 replacement of the old behavior.
SimpleRequestStrategy accepts an optional pre-configured Session if the caller
needs to do more specific set-up, e.g. of headers, etc.

RetryingRequestStrategy relies on `tenacity.Retrying`, from tenacity, a
well-established, widely trusted, and batteries-included function-retry library.
The caller must provide a configured Retrying object to control the retry
behavior on the requests.

RateLimitRetryingRequestStrategy subclasses RetryingRequestStrategy and provides
a Retrying object that reimplements the built-in retry behavior of the official
Airtable JS library, airtable.js. Requests to Airtable that hit the rate limit
of 5 request/second, are rejected with a 429 status code, as are all subsequent
requests for 30 seconds. airtable.js accommodates this with a jittered
exponential backoff starting at 30s when it receives a 429, and that behavior is
implemented here.

---

## To Do

- [x] Add tests
- [x] Better type checking on request() to assert expected return shapes

### DIFF
--- a/pyairtable/api/abstract.py
+++ b/pyairtable/api/abstract.py
@@ -1,13 +1,12 @@
 import abc
 from functools import lru_cache
 import posixpath
-from typing import List, Optional, Tuple
+from typing import List, Optional, Tuple, Type, Union
 import time
 from urllib.parse import quote
 
-import requests
-
 from .params import to_params_dict
+from ..request_strategies import RequestStrategy, SimpleRequestStrategy, Headers
 
 
 class ApiAbstract(metaclass=abc.ABCMeta):
@@ -17,28 +16,36 @@ class ApiAbstract(metaclass=abc.ABCMeta):
     API_URL = posixpath.join(API_BASE_URL, VERSION)
     MAX_RECORDS_PER_REQUEST = 10
 
-    session: requests.Session
-    tiemout: Optional[Tuple[int, int]]
+    _request_strategy: Optional[RequestStrategy] = None
 
-    def __init__(self, api_key: str, timeout=None):
-        session = requests.Session()
-        self.session = session
+    def __init__(
+        self,
+        api_key: str,
+        timeout: Optional[Tuple[int, int]] = None,
+        request_strategy: Optional[
+            Union[Type[RequestStrategy], RequestStrategy]
+        ] = None,
+    ):
+        if request_strategy is not None:
+            self.request_strategy = request_strategy
         self.timeout = timeout
         self.api_key = api_key
 
     @property
-    def api_key(self) -> str:
-        """Returns the Airtable API Key"""
-        return self._api_key
+    def request_strategy(self) -> RequestStrategy:
+        if self._request_strategy is None:
+            self._request_strategy = SimpleRequestStrategy()
+        return self._request_strategy
 
-    @api_key.setter
-    def api_key(self, value):
-        """Returns the Airtable API Key"""
-        self._update_api_key(value)
-        self._api_key = value
+    @request_strategy.setter
+    def request_strategy(
+        self, _request_strategy: Union[Type[RequestStrategy], RequestStrategy]
+    ) -> None:
+        self._request_strategy = RequestStrategy.initialize(_request_strategy)
 
-    def _update_api_key(self, api_key: str) -> None:
-        self.session.headers.update({"Authorization": "Bearer {}".format(api_key)})
+    @property
+    def auth_headers(self) -> Headers:
+        return {"Authorization": "Bearer {}".format(self.api_key)}
 
     @lru_cache()
     def get_table_url(self, base_id: str, table_name: str):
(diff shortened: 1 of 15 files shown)


gtalarico:
Thanks for sharing this @jskrzypek 
It looks great to me so far 👌

jskrzypek:
@gtalarico I think the only thing still missing is the docs. Other than that I am not sure exactly how to start providing integration tests for the retrying strategy, as I'd need to violate the rate limit, which seems like a bad idea.

jskrzypek:
Fwiw, I'm planning to start using this from my fork in some in-development code for now, so the code will get a full workthrough

jskrzypek:
Ok I think I figured out the rst to handle the docs the right way now.

jskrzypek:
@gtalarico This is ready to review when you have time... If you prefer / think it would be helpful I'm happy to schedule a video chat sometime for us to go over the PR together.

pyairtable/api/abstract.py
```diff
@@ -17,28 +16,36 @@ class ApiAbstract(metaclass=abc.ABCMeta):
...
 
-    def __init__(self, api_key: str, timeout=None):
-        session = requests.Session()
-        self.session = session
+    def __init__(
+        self,
+        api_key: str,
+        timeout: Optional[Tuple[int, int]] = None,
+        request_strategy: Optional[
+            Union[Type[RequestStrategy], RequestStrategy]
```
gtalarico, line 26: Can you explain the type here?

pyairtable/api/abstract.py
```diff
@@ -17,28 +16,36 @@ class ApiAbstract(metaclass=abc.ABCMeta):
...
+        if request_strategy is not None:
+            self.request_strategy = request_strategy
         self.timeout = timeout
         self.api_key = api_key
 
     @property
-    def api_key(self) -> str:
-        """Returns the Airtable API Key"""
-        return self._api_key
+    def request_strategy(self) -> RequestStrategy:
```
gtalarico, line 35: What you think of simplifying this to:

```
def __init__(..., request_strategy=...):
    ...
    self.request_strategy = request_strategy or SimpleRequestStrategy()
    self.timeout = timeout
    self.api_key = api_key
```

gtalarico (approved):
Very Happy with the PR overall! Let's work together to tie it all up with docs and examples

gtalarico:
For docs, maybe a new section under `api.rst` - example below.
Then run `make docs` to preview.


```
Request Strategies
*******************

[ Overview, explain default behavior where strategy is None ]
[ Simple Example ]

Custom Request Strategy
--------------------------
[ Example ]

```

gtalarico:
> @gtalarico This is ready to review when you have time... If you prefer / think it would be helpful I'm happy to schedule a video chat sometime for us to go over the PR together.

Happy to as well. What timezone are you on?

jskrzypek:
I'm on EST (New York)

pyairtable/request_strategies/simple.py
```diff
@@ -0,0 +1,52 @@
...
+        self,
+        session: Optional[Session] = None,
+    ):
+        """
+        Initialize a simple strategy that just proxies requests to its Session.
+
+        Args:
+            session(``Session``, optional): |arg_session|
+        """
+        self.session = session if session is not None else Session()
```
gtalarico, line 28: `self.session = session or Session()` should do it?

pyairtable/request_strategies/retrying.py
```diff
@@ -0,0 +1,252 @@
...
+                "keyword arguments! Got neither!"
+            )
+        self.retrying = self.__class__.make_retrying(**kwargs)
+
+    @property
+    def retrying(self) -> Retrying:
+        """Return the tenacity.Retrying object that wraps _request."""
+        return self._retrying
+
+    if TENACITY_FOUND:
```
gtalarico, line 135: wow TIL you can have functional definitions inside conditional statements

pyairtable/request_strategies/retrying.py
```diff
@@ -0,0 +1,252 @@
+from typing import Any, Optional, overload
```
gtalarico, line 1: Is there a way to simplify this (module)? 
Given the size of the library, this is sizable addition and I am concerned about maintaining a lot of complex logic in this module, some of which I am still trying to wrap my head around

gtalarico (commented):
Comments

jskrzypek:
Hi @gtalarico this week was a bit of a hell week for us, but I'm hoping next week is better... Let me get through the weekend and I'll respond to your comments and ping you to find a time that will work for us to go over this stuff

ebsaral:
I was about to use `pyairtable` but I am sure I will also hit the rate limits issue. I am looking forward to seeing this PR merged. Until then, I will be using rest api. :) 

Thank you for the great work, both of you!

mesozoic:
Hi there, I'm going through and trying to clean up old PRs in this repo. Now that there is an option to pass a retry parameter, this issue seems resolved, so I'm going to close this pull request. If there's additional functionality you'd like to see that's not in main, please feel free to submit another patch.

PR #144, opened 2022-03-02
Parameterize max records per request

lucas-simonelli:
I'm pushing a lot of records to a table (~10k), so I'd rather try more than 10 per request in the batch creation/updates. This PR parameterized the MAX_RECORD_PER_REQUEST arg, defaulting to 10

### DIFF
--- a/pyairtable/api/abstract.py
+++ b/pyairtable/api/abstract.py
@@ -19,7 +19,7 @@ class ApiAbstract(metaclass=abc.ABCMeta):
     API_URL = posixpath.join(API_BASE_URL, VERSION)
 
     session: requests.Session
-    tiemout: Optional[Tuple[int, int]]
+    timeout: Optional[Tuple[int, int]]
 
     def __init__(self, api_key: str, max_records_per_request: int, timeout=None):
         session = requests.Session()
--- a/pyairtable/api/abstract.py
+++ b/pyairtable/api/abstract.py
@@ -66,7 +66,7 @@ class ApiAbstract(metaclass=abc.ABCMeta):
     def _chunk(self, iterable, chunk_size):
         """Break iterable into chunks"""
         for i in range(0, len(iterable), chunk_size):
-            yield iterable[i: i + chunk_size]
+            yield iterable[i : i + chunk_size]
 
     def _build_batch_record_objects(self, records):
         return [{"fields": record} for record in records]
--- a/pyairtable/api/abstract.py
+++ b/pyairtable/api/abstract.py
@@ -9,7 +9,7 @@ import requests
 
 from .params import to_params_dict
 
-MAX_RECORDS_PER_REQUEST = 10
+DEFAULT_MAX_RECORDS_PER_REQUEST = 10
 
 
 class ApiAbstract(metaclass=abc.ABCMeta):
@@ -21,7 +21,8 @@ class ApiAbstract(metaclass=abc.ABCMeta):
     session: requests.Session
     timeout: Optional[Tuple[int, int]]
 
-    def __init__(self, api_key: str, max_records_per_request: int, timeout=None):
+    def __init__(self, api_key: str, timeout: Optional[int]=None,
+                 max_records_per_request: Optional[int]=DEFAULT_MAX_RECORDS_PER_REQUEST):
         session = requests.Session()
         self.session = session
         self.timeout = timeout
--- a/pyairtable/api/abstract.py
+++ b/pyairtable/api/abstract.py
@@ -9,22 +9,24 @@ import requests
 
 from .params import to_params_dict
 
+MAX_RECORDS_PER_REQUEST = 10
+
 
 class ApiAbstract(metaclass=abc.ABCMeta):
     VERSION = "v0"
     API_BASE_URL = "https://api.airtable.com/"
     API_LIMIT = 1.0 / 5  # 5 per second
     API_URL = posixpath.join(API_BASE_URL, VERSION)
-    MAX_RECORDS_PER_REQUEST = 10
 
     session: requests.Session
     tiemout: Optional[Tuple[int, int]]
 
-    def __init__(self, api_key: str, timeout=None):
+    def __init__(self, api_key: str, max_records_per_request, timeout=None):
         session = requests.Session()
         self.session = session
         self.timeout = timeout
         self.api_key = api_key
+        self.max_records_per_request = max_records_per_request
 
     @property
     def api_key(self) -> str:
@@ -64,7 +66,7 @@ class ApiAbstract(metaclass=abc.ABCMeta):
     def _chunk(self, iterable, chunk_size):
         """Break iterable into chunks"""
         for i in range(0, len(iterable), chunk_size):
-            yield iterable[i : i + chunk_size]
+            yield iterable[i: i + chunk_size]
 
     def _build_batch_record_objects(self, records):
         return [{"fields": record} for record in records]
@@ -143,7 +145,7 @@ class ApiAbstract(metaclass=abc.ABCMeta):
 
         table_url = self.get_table_url(base_id, table_name)
         inserted_records = []
-        for chunk in self._chunk(records, self.MAX_RECORDS_PER_REQUEST):
+        for chunk in self._chunk(records, self.max_records_per_request):
             new_records = self._build_batch_record_objects(chunk)
             response = self._request(
                 "post",
@@ -181,7 +183,7 @@ class ApiAbstract(metaclass=abc.ABCMeta):
         updated_records = []
         table_url = self.get_table_url(base_id, table_name)
         method = "put" if replace else "patch"
-        for records in self._chunk(records, self.MAX_RECORDS_PER_REQUEST):
+        for records in self._chunk(records, self.max_records_per_request):
             chunk_records = [{"id": x["id"], "fields": x["fields"]} for x in records]
             response = self._request(
                 method,
@@ -201,7 +203,7 @@ class ApiAbstract(metaclass=abc.ABCMeta):
     ) -> List[dict]:
         deleted_records = []
         table_url = self.get_table_url(base_id, table_name)
-        for record_ids in self._chunk(record_ids, self.MAX_RECORDS_PER_REQUEST):
+        for record_ids in self._chunk(record_ids, self.max_records_per_request):
             delete_results = self._request(
                 "delete", table_url, params={"records[]": record_ids}
             )
--- a/pyairtable/api/abstract.py
+++ b/pyairtable/api/abstract.py
@@ -21,7 +21,7 @@ class ApiAbstract(metaclass=abc.ABCMeta):
     session: requests.Session
     tiemout: Optional[Tuple[int, int]]
 
-    def __init__(self, api_key: str, max_records_per_request, timeout=None):
+    def __init__(self, api_key: str, max_records_per_request: int, timeout=None):
         session = requests.Session()
         self.session = session
         self.timeout = timeout
--- a/pyairtable/api/api.py
+++ b/pyairtable/api/api.py
@@ -18,7 +18,12 @@ class Api(ApiAbstract):
         >>> api.all('base_id', 'table_name')
     """
 
-    def __init__(self, api_key: str, timeout=None, max_records_per_request=MAX_RECORDS_PER_REQUEST):
+    def __init__(
+        self,
+        api_key: str,
+        timeout=None,
+        max_records_per_request=MAX_RECORDS_PER_REQUEST,
+    ):
         """
 
         Args:
@@ -28,7 +33,9 @@ class Api(ApiAbstract):
             timeout(``Tuple``): |arg_timeout|
             max_records_per_request: |arg_max_records_per_request|
         """
-        super().__init__(api_key, max_records_per_request=max_records_per_request, timeout=timeout)
+        super().__init__(
+            api_key, max_records_per_request=max_records_per_request, timeout=timeout
+        )
 
     def get_table(self, base_id: str, table_name: str) -> "Table":
         """
--- a/pyairtable/api/api.py
+++ b/pyairtable/api/api.py
@@ -1,6 +1,6 @@
 from typing import List
 
-from .abstract import ApiAbstract, MAX_RECORDS_PER_REQUEST
+from .abstract import ApiAbstract, DEFAULT_MAX_RECORDS_PER_REQUEST
 
 
 class Api(ApiAbstract):
@@ -22,7 +22,7 @@ class Api(ApiAbstract):
         self,
         api_key: str,
         timeout=None,
-        max_records_per_request=MAX_RECORDS_PER_REQUEST,
+        max_records_per_request=DEFAULT_MAX_RECORDS_PER_REQUEST,
     ):
         """
 
--- a/pyairtable/api/api.py
+++ b/pyairtable/api/api.py
@@ -23,10 +23,10 @@ class Api(ApiAbstract):
 
         Args:
             api_key: |arg_api_key|
+            max_records_per_request: |arg_max_records_per_request|
 
         Keyword Args:
             timeout(``Tuple``): |arg_timeout|
-            max_records_per_request: |arg_max_records_per_request|
         """
         super().__init__(api_key, max_records_per_request=max_records_per_request, timeout=timeout)
 
--- a/pyairtable/api/api.py
+++ b/pyairtable/api/api.py
@@ -1,5 +1,6 @@
 from typing import List
-from .abstract import ApiAbstract
+
+from .abstract import ApiAbstract, MAX_RECORDS_PER_REQUEST
 
 
 class Api(ApiAbstract):
@@ -17,7 +18,7 @@ class Api(ApiAbstract):
         >>> api.all('base_id', 'table_name')
     """
 
-    def __init__(self, api_key: str, timeout=None):
+    def __init__(self, api_key: str, max_records_per_request=MAX_RECORDS_PER_REQUEST, timeout=None):
         """
 
         Args:
@@ -25,9 +26,9 @@ class Api(ApiAbstract):
 
         Keyword Args:
             timeout(``Tuple``): |arg_timeout|
-
+            max_records_per_request: |arg_max_records_per_request|
         """
-        super().__init__(api_key, timeout=timeout)
+        super().__init__(api_key, max_records_per_request=max_records_per_request, timeout=timeout)
 
     def get_table(self, base_id: str, table_name: str) -> "Table":
         """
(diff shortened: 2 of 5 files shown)


pyairtable/api/abstract.py
```diff
@@ -9,22 +9,24 @@
...
     API_BASE_URL = "https://api.airtable.com/"
     API_LIMIT = 1.0 / 5  # 5 per second
     API_URL = posixpath.join(API_BASE_URL, VERSION)
-    MAX_RECORDS_PER_REQUEST = 10
 
     session: requests.Session
     tiemout: Optional[Tuple[int, int]]
 
-    def __init__(self, api_key: str, timeout=None):
+    def __init__(self, api_key: str, max_records_per_request, timeout=None):
```
kkom, line 24: You'll need an argument typehint here!

pyairtable/api/api.py
```diff
@@ -17,17 +18,17 @@ class Api(ApiAbstract):
...
+    def __init__(self, api_key: str, max_records_per_request=MAX_RECORDS_PER_REQUEST, timeout=None):
         """
 
         Args:
             api_key: |arg_api_key|
 
         Keyword Args:
             timeout(``Tuple``): |arg_timeout|
-
+            max_records_per_request: |arg_max_records_per_request|
```
kkom, line 29: Also, the order in the docstring is different than of the actual arguments

pyairtable/api/abstract.py
```diff
@@ -9,22 +9,24 @@
...
     API_BASE_URL = "https://api.airtable.com/"
     API_LIMIT = 1.0 / 5  # 5 per second
     API_URL = posixpath.join(API_BASE_URL, VERSION)
-    MAX_RECORDS_PER_REQUEST = 10
 
     session: requests.Session
     tiemout: Optional[Tuple[int, int]]
 
-    def __init__(self, api_key: str, timeout=None):
+    def __init__(self, api_key: str, max_records_per_request: int, timeout=None):
```
kkom, line 24: I'm on mobile, so that's why I just send random notes here and there.

Is it possible that someone uses those arguments in a positional, rather than keyword, manner?

Then, after an upgrade, some users will have their timeout interpreted as max_records_per_request?

lucas-simonelli: timeout is a named keyword argument, and can't be used without passing the name. For better bw compat though, I think i should move it to the end

kkom: are you sure?

```python
>>> def foo(a: int = 1, c: int = 3) -> None:
...     print(f"{a=}")
...     print(f"{c=}")
... 
>>> def bar(a: int = 1, b: int = 2, c: int = 3) -> None:
...     print(f"{a=}")
...     print(f"{b=}")
...     print(f"{c=}")
... 
>>> foo(101, 102)
a=101
c=102
>>> bar(101, 102)
a=101
b=102
c=3
```

kkom: This is just a runtime example of course, but not everyone uses static analysis tools in Python. Either way, I think developers are not required to name the arguments - even with strict type checking.

kkom: Yeah, there isn't another way I think. @gtalarico – thoughts?

pyairtable/api/abstract.py
```diff
@@ -64,7 +66,7 @@ def _options_to_params(self, **options):
     def _chunk(self, iterable, chunk_size):
         """Break iterable into chunks"""
         for i in range(0, len(iterable), chunk_size):
-            yield iterable[i : i + chunk_size]
+            yield iterable[i: i + chunk_size]
```
kkom, line 69: as a nitpick you modified some lines with just whitespace changes, sometimes even ones against Python's style guidelines (there are few more places with unnecessary double spaces, no whitespace at all, etc.)

lucas-simonelli: I made my changes [PEP8](https://www.python.org/dev/peps/pep-0008/) compliant

kkom: i actually don't think so!

Screenshot 2022-03-04 at 18 28 43

either way – doesn't matter, you should run this to autoformat the entire project:

https://github.com/gtalarico/pyairtable/blob/bb13d4c06e3917682c1900afbb14f88ba65adc5a/Makefile#L41-L42

I'm surprised that CI didn't include a check that the files are formatted! 🤔

lucas-simonelli: dayum, my IDE was lying to me there, I guess i should not try the built in formatter :(

pyairtable/api/abstract.py
gtalarico: Agree it should go last for compat.

I forgot what this is called, but maybe we use the `*` separator to require named arg? 
I think that's probably best practice for a lib. 
Maybe we should required named args for all kwargs...?

```
>>> def test(x, y="b", *, z="c"):
...  print(x)
...  print(y)
...  print(z)
...
>>> test(1,2)
1
2
c
>>> test(1, 2, 3)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: test() takes from 1 to 2 positional arguments but 3 were given
>>> test(1, 2, z=3)
1
2
3
```

pyairtable/api/abstract.py
gtalarico: It is supposed to run flake8 and black in CI but I think I forgot to configure it ! 🙈 
The CI workflow is even called `test, lint, deploy` :)
https://github.com/gtalarico/pyairtable/blob/main/.github/workflows/test_lint_deploy.yml
https://github.com/gtalarico/pyairtable/blob/main/tox.ini#L18

pyairtable/api/api.py
```diff
@@ -17,17 +18,24 @@ class Api(ApiAbstract):
         >>> api.all('base_id', 'table_name')
     """
 
-    def __init__(self, api_key: str, timeout=None):
+    def __init__(
+        self,
+        api_key: str,
+        timeout=None,
+        max_records_per_request=MAX_RECORDS_PER_REQUEST,
```
gtalarico, line 25: WDYT about `max_records_per_request= ApiAbstract.DEFAULT_MAX_RECORDS_PER_REQUEST` instead of importing the constant?

pyairtable/api/abstract.py
```diff
@@ -9,22 +9,24 @@
...
 
 class ApiAbstract(metaclass=abc.ABCMeta):
     VERSION = "v0"
     API_BASE_URL = "https://api.airtable.com/"
     API_LIMIT = 1.0 / 5  # 5 per second
     API_URL = posixpath.join(API_BASE_URL, VERSION)
-    MAX_RECORDS_PER_REQUEST = 10
 
     session: requests.Session
     tiemout: Optional[Tuple[int, int]]
```
gtalarico, line 22: typo in the lib? 🙈

pyairtable/api/abstract.py
gtalarico: Either way, for now this should be probably be a kwarg not positional. 
#1 because it seems like it should be an optional parameter with a sane default, but #2 to not break existing code on upgrade.
I think I would probably do something like this 👇 ?
```
def __init__(self, api_key: str, timeout: Optional[int]=None, max_records_per_request: Optional[int]):
   max_records_per_request = max_records_per_request or self.DEFAULT_MAX_RECORDS_PER_REQUEST
   ...
```

pyairtable/api/api.py
lucas-simonelli: Hmm it's a module variable, so it can't be imported like that

pyairtable/api/abstract.py
lucas-simonelli: yup, fixed!

pyairtable/api/api.py
kkom: I think it can! As long as you do `import ApiAbstract`

lucas-simonelli: It's a module level variable, not a class variable, so it could be abstract.DEFAULT_MAX_RECORDS_PER_REQUEST I guess

kkom:
Hey @gtalarico – I will actually need this PR also in a side project of mine! :) Would mind taking a look when you have a moment?

gtalarico:
@kkom
There are still some unresolved issues above. 
And I will still need to tag and ship after, so unfortunately this might not be ready for when you needed!

However, you can install and add dep from a branch/commit, so hopefully you can use that as a temporary fix 👇 
https://stackoverflow.com/questions/16584552/how-to-state-in-requirements-txt-a-direct-github-source
https://stackoverflow.com/questions/20101834/pip-install-from-git-repo-branch

kkom:
No worries, I can wait a little. It turned out that my side project is fast enough for now anyway, so there is no terrible rush. Thanks @gtalarico !

pyairtable/api/abstract.py
```diff
@@ -9,22 +9,25 @@
...
     API_URL = posixpath.join(API_BASE_URL, VERSION)
-    MAX_RECORDS_PER_REQUEST = 10
 
     session: requests.Session
-    tiemout: Optional[Tuple[int, int]]
+    timeout: Optional[Tuple[int, int]]
 
-    def __init__(self, api_key: str, timeout=None):
+    def __init__(self, api_key: str, timeout: Optional[int]=None,
+                 max_records_per_request: Optional[int]=DEFAULT_MAX_RECORDS_PER_REQUEST):
```
jskrzypek, line 25: This isn't black formatted. You ought to install the packages in `requirements-dev.txt` and then run 
```sh
black .
````

File: pyairtable/api/api.py [hunk 8]
```diff
@@ -17,17 +18,24 @@ class Api(ApiAbstract):
...
             api_key: |arg_api_key|
 
         Keyword Args:
             timeout(``Tuple``): |arg_timeout|
-
+            max_records_per_request: |arg_max_records_per_request|
         """
-        super().__init__(api_key, timeout=timeout)
+        super().__init__(
+            api_key, max_records_per_request=max_records_per_request, timeout=timeout
```
jskrzypek, line 37: This should also be the reversed order...

pyairtable/api/abstract.py
jskrzypek, line 25: Perhaps `timeout` and `max_records_per_request` should be explicitly keyword-only?
This is already the case in the `Table`'s __init__ method, and following through to here would provide a clearer extension path and also cleaner/DRYer code in the subclasses.
```suggestion
    def __init__(
        self,
        api_key: str,
        *, 
        timeout: Optional[int]=None,
        max_records_per_request: Optional[int]=DEFAULT_MAX_RECORDS_PER_REQUEST
    ):
```

pyairtable/api/api.py
```diff
@@ -17,17 +18,24 @@ class Api(ApiAbstract):
         >>> api.all('base_id', 'table_name')
     """
 
-    def __init__(self, api_key: str, timeout=None):
+    def __init__(
+        self,
+        api_key: str,
+        timeout=None,
+        max_records_per_request=DEFAULT_MAX_RECORDS_PER_REQUEST,
+    ):
```
jskrzypek, line 26: This feels like an awkward pattern or code-smell, since it requires importing the default value and using it as a parameter default, which then must be passed and will override the default provided by the superclass.

A better pattern might be to just move the `*` explicit keyword-only argument separator to the `AbstractApi.__init__` from the `Table.__init__` and make both `timeout` and `max_records_per_request` into keyword-only arguments (this is my suggestion above) and then here (as well as `Base` & `Table`) just accept and pass a `**kwargs` dictionary. 

```suggestion
    def __init__(self, api_key: str, **kwargs: Any):
```
_NB, don't forget to add the import for `Any` from `typing` above_

pyairtable/api/base.py
```diff
@@ -15,18 +15,27 @@ class Base(ApiAbstract):
...
     base_id: str
 
-    def __init__(self, api_key: str, base_id: str, timeout=None):
+    def __init__(
+        self,
+        api_key: str,
+        base_id: str,
+        timeout=None,
+        max_records_per_request=DEFAULT_MAX_RECORDS_PER_REQUEST,
+    ):
```
jskrzypek, line 24: See my reasoning above for `Api`

```suggestion
    def __init__(self, api_key: str, base_id: str, **kwargs: Any):
```
_NB, don't forget to add the import for `Any` from `typing` above_

pyairtable/api/api.py
```diff
@@ -17,17 +18,24 @@ class Api(ApiAbstract):
...
         """
 
         Args:
             api_key: |arg_api_key|
 
         Keyword Args:
             timeout(``Tuple``): |arg_timeout|
-
+            max_records_per_request: |arg_max_records_per_request|
```
jskrzypek, line 34: Assuming my suggestions are accepted for the keyword-only argument changes, this snippet of docstring can also become an RST replacement for the keyword-only args accepted by the `AbstractApi` class, and reused in all 3 subclasses.

pyairtable/api/api.py
```diff
@@ -17,17 +18,24 @@ class Api(ApiAbstract):
...
         """
-        super().__init__(api_key, timeout=timeout)
+        super().__init__(
+            api_key, max_records_per_request=max_records_per_request, timeout=timeout
+        )
```
jskrzypek, line 38: Carrying through my suggestions for keyword-only arguments...
```suggestion
        super().__init__(api_key, **kwargs)
```

pyairtable/api/table.py
```diff
@@ -16,7 +16,15 @@ class Table(ApiAbstract):
...
-    def __init__(self, api_key: str, base_id: str, table_name: str, *, timeout=None):
+    def __init__(
+        self,
+        api_key: str,
+        base_id: str,
+        table_name: str,
+        *,
+        timeout=None,
+        max_records_per_request=DEFAULT_MAX_RECORDS_PER_REQUEST,
+    ):
```
jskrzypek, line 27: See my reasoning above for `Api`

```suggestion
    def __init__(self, api_key: str, base_id: str, table_name: str, **kwargs: Any):
```
_NB, don't forget to add the import for `Any` from `typing` above_

pyairtable/api/base.py
```diff
@@ -15,18 +15,27 @@ class Base(ApiAbstract):
...
         Keyword Args:
             timeout(``Tuple``): |arg_timeout|
+            max_records_per_request: |arg_max_records_per_request|
         """
 
         self.base_id = base_id
-        super().__init__(api_key, timeout=timeout)
+        super().__init__(
+            api_key, max_records_per_request=max_records_per_request, timeout=timeout
+        )
```
jskrzypek, line 38: Carrying through my suggestions for keyword-only arguments...
```suggestion
        super().__init__(api_key, **kwargs)
```

pyairtable/api/table.py
```diff
@@ -25,10 +33,13 @@ def __init__(self, api_key: str, base_id: str, table_name: str, *, timeout=None)
...
         Keyword Args:
             timeout(``Tuple``): |arg_timeout|
+            max_records_per_request: |arg_max_records_per_request|
         """
         self.base_id = base_id
         self.table_name = table_name
-        super().__init__(api_key, timeout=timeout)
+        super().__init__(
+            api_key, max_records_per_request=max_records_per_request, timeout=timeout
+        )
```
jskrzypek, line 42: Carrying through my suggestions for keyword-only arguments...
```suggestion
        super().__init__(api_key, **kwargs)
```

pyairtable/api/table.py
```diff
@@ -1,5 +1,5 @@
 from typing import List
-from .abstract import ApiAbstract
+from .abstract import ApiAbstract, DEFAULT_MAX_RECORDS_PER_REQUEST
```
jskrzypek, line 2: Accepting my suggestions for keyword-only arguments means this import would be unnecessary
```suggestion
from .abstract import ApiAbstract
```

pyairtable/api/base.py
```diff
@@ -1,6 +1,6 @@
 from typing import List
 
-from .abstract import ApiAbstract
+from .abstract import ApiAbstract, DEFAULT_MAX_RECORDS_PER_REQUEST
```
jskrzypek, line 3: Accepting my suggestions for keyword-only arguments means this import would be unnecessary
```suggestion
from .abstract import ApiAbstract
```

pyairtable/api/api.py
```diff
@@ -1,5 +1,6 @@
 from typing import List
-from .abstract import ApiAbstract
+
+from .abstract import ApiAbstract, DEFAULT_MAX_RECORDS_PER_REQUEST
```
jskrzypek, line 3: Accepting my suggestions for keyword-only arguments means this import would be unnecessary
```suggestion

from .abstract import ApiAbstract
```

pyairtable/api/abstract.py
```diff
@@ -9,22 +9,25 @@
...
 class ApiAbstract(metaclass=abc.ABCMeta):
     VERSION = "v0"
     API_BASE_URL = "https://api.airtable.com/"
     API_LIMIT = 1.0 / 5  # 5 per second
     API_URL = posixpath.join(API_BASE_URL, VERSION)
-    MAX_RECORDS_PER_REQUEST = 10
 
     session: requests.Session
-    tiemout: Optional[Tuple[int, int]]
+    timeout: Optional[Tuple[int, int]]
```
jskrzypek, line 22: Good catch :)

jskrzypek (commented):
Nice contribution! I think this would be a good opportunity to make the `__init__` methods and docs a little DRYer, but YMMV

pyairtable/api/abstract.py
gtalarico: Related:
https://github.com/gtalarico/pyairtable/pull/144#discussion_r826602036

gtalarico: +1 I also prefer this and wish I had added from the beginning 
However, we should consider adding this would be a breaking change

jskrzypek: Good point. I doubt anyone is passing the timeout argument as a positional parameter, but better to be nice about it :P

gtalarico: created a fresh CI to test CI, and seem to catch formatting, not sure why this CI steps not running on this PR:
https://github.com/gtalarico/pyairtable/pull/151

jskrzypek: @gtalarico The branch needs to be rebased on master to have the changes to the .github folder picked up.

mesozoic:
Hi there, I'm going through old PRs, and I'm wondering if you've found a way to make this sort of thing work?.When I try to pass more than 10 records at a time to one of the batch APIs, I get a 422 error:

```json
{
  "type": "INVALID_RECORDS",
  "message": "A maximum of 10 records can be created per request but you have provided 100."
}
```

Since more than 10 records is explicitly not supported and does not appear to work, can I close this PR?

lucas-simonelli:
Might be that they changed the API limit after all this time. Closing it then!

PR #148, opened 2022-03-21
feat: add OR formula

a5r0n:


gtalarico:
Thanks. We will need to add docs as well

a5r0n:
maybe we want to add something like `match_any` & `match_all` instead of only `match`

gtalarico:
> maybe we want to add something like match_any & match_all instead of only match
That's a good idea.
I believe `match` is currently a match all.

two options ideas:
1. e can just add a new `match_any`?
2. we can modify `match(dict)` to `def match(dict, all=True)` where `all=False` is an `OR`, and `all=True` = `AND`

PR #149, opened 2022-04-05
add support for PercentField in orm

prenit-wankhede:


gtalarico:
I have a question @prenit-wankhede , not sure if you know - 
1. does airtable return percentages as 0-1 or 0-100? 
2. does it return floats or Ints?

aleafa:
@gtalarico Airtable returns (and takes) percentages as a decimal value between 0 and 1 ([optionally including negative values](https://support.airtable.com/docs/percent-field-type)). For the `PercentField` subclass is to support `int` values or `float` values outside that range, shouldn't it convert them to an acceptable format?

@prenit-wankhede would mathematical operations with `PercentField` values be easier if it consistently returned a `float` between -1 and 1?

PR #151, opened 2022-04-11
Fix CI Linting

gtalarico:
PR that introduces `black` formatting violation to test if CI is properly linting


gtalarico:
Failed as expected:
https://github.com/gtalarico/pyairtable/runs/5976947909?check_suite_focus=true#step:6:50

PR #153, opened 2022-04-19
fix: key error in doc

nikiljos:
I was going through the docs and can across this error copy pasting from it
image

The parameter in the doc is `maxRecords`(which is actually the airtable api parameter). What worked for me instead is `max_records`. 

I've made this small PR in order to fix the doc..... And thanks a lot for this awesome package.... It helped me a lot in my recent project....


gtalarico:
Thanks. Previous versions supported both.

PR #157, opened 2022-05-06
Add returnFieldsByFieldId parameter

bfortuner:
AirTable REST API allows a nice parameter to use field_ids instead of field names. This makes programming against user-edited tables more reliable, since the field ids don't change, even if field names do.

```
returnFieldsByFieldId
boolean
optional
An optional boolean value that lets you return field objects where the key is the field id.

This defaults to false, which returns field objects where the key is the field name.
```

### DIFF
--- a/pyairtable/api/params.py
+++ b/pyairtable/api/params.py
@@ -85,6 +85,8 @@ def to_params_dict(param_name: str, value: Any):
         return {"timeZone": value}
     elif param_name == "user_locale":
         return {"userLocale": value}
+    elif param_name == "returnFieldsByFieldId":
+        return {"returnFieldsByFieldId": value}
     elif param_name == "sort":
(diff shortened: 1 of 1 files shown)


inoatech:
Perhaps use return_fields_by_field_id instead of returnFieldsByFieldId for consistency?

robbieaverill:
+1 on this, Airtable API docs recommend using the ID over field names:

> Field names and field ids can be used interchangeably. Using field ids means field name changes do not require modifications to your API request. We recommend using field ids over field names where possible, to reduce modifications to your API request if the user changes the field name later.

cc @gtalarico

robbieaverill:
I will make a new PR with this commit cherry-picked on Monday if OP doesn't beat me to it by then 🙂

robbieaverill:
This is available in v1.2.0 now

PR #159, opened 2022-06-09
Track changes made to ORM fields since fetch/save

mesozoic:
I have some use cases that require me to keep track of which fields I've changed. This branch adds a new public property to each model, `Model.changes`, which will record any changes made to the model's fields since the last time it was fetched or persisted. The implementation relies on each field telling the model instance about a change as it happens.

This could probably be a lot more robust and featureful, but I didn't want to overcomplicate things before writing a few tests for the basics and opening a branch.

### DIFF
--- a/pyairtable/orm/fields.py
+++ b/pyairtable/orm/fields.py
@@ -96,7 +96,10 @@ class Field(metaclass=abc.ABCMeta):
             instance._fields = {}
         if self.validate_type:
             self.valid_or_raise(value)
+        previous = instance._fields.get(self.field_name, None)
         instance._fields[self.field_name] = value
+        if hasattr(instance, "_changes"):
+            instance._changes.observe(self.attribute_name, previous, value)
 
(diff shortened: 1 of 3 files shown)


fbertoia:
Is this something like this you are looking for ? https://docs.djangoproject.com/en/4.1/ref/models/instances/#django.db.models.Model.refresh_from_db

I suggest the same keyword then. 😀

mesozoic:
> Is this something like this you are looking for ? https://docs.djangoproject.com/en/4.1/ref/models/instances/#django.db.models.Model.refresh_from_db

Not exactly. I'm more thinking of "dirty fields", similar to the [django-dirtyfields](https://django-dirtyfields.readthedocs.io/en/stable/quickstart.html#usage) plugin. I could've mimicked that API more closely but wanted to get some feedback on the idea before building out any more functionality.

mesozoic:
This is my own branch and it's kinda outdated. I'll close for now and we can come back to this later.

PR #178, opened 2022-09-09
Flagging computed fields in ORM

larsakerson:
Add flag for computed fields in `orm.fields.Field` and use this attribute to resolve error on record updates in `orm.model.Model.save()`


gtalarico:
Thanks for the quick turn around.

This looks good to me. 
Just need a few more things to make it ready to ship:

* Can you add a unit test in the orm test suite?
* maybe a note in the orm docs?

NicoHood:
Wouldnt it make sense to rename this parameter to something else? I was thinking about this, as I need to download images via airtable:
https://github.com/gtalarico/pyairtable/issues/179

Now if we want to add this image implementation, I think it could be tricky to get it right to support download and upload. I think for the first run a download-only option would make sense, and then this PR comes into play:

In this case the image fields are not computed, but we rather want to treat it `read-only`. So maybe a more generic name would be `read_only` in this case?

aleafa:
Your suggestion makes sense to me, @NicoHood. I've updated the parameter name to `read_only`.

@gtalarico I'm not very familiar with pytest 😬 but I added a test and updated the test suite for this feature and they pass. Might be worth another set of eyes on the code, though.

pyairtable/orm/model.py
```diff
@@ -209,7 +209,8 @@ def save(self) -> bool:
         """
         table = self.get_table()
         record = self.to_record()
-        fields = record["fields"]
+        _fndm = self._field_name_descriptor_map()
```
NicoHood, line 212: I personally avoid such abbreviations in variables. Of course, you can the the name on the right hand side, but it leads to abuse in further commits. Also I'd like to have a comment on what the loop does. Maybe it makes sense to split this one liner into multiple lines but explain it better.

tests/test_orm_fields.py
```diff
@@ -27,6 +27,15 @@ class T:
     assert t.email == "x@x.com"
 
 
+def test_field_read_only():
+    class T:
+        name = f.Field("Name", read_only=True)
+
+    t = T()
+    t.name = "y"
+    assert t.name is None
```
NicoHood, line 36: what does this test do? shouldnt setting the same rather throw an exception?

tests/test_orm.py
```diff
@@ -52,6 +52,7 @@ class Contact(Model):
         is_registered = f.CheckboxField("Registered")
         link = f.LinkField("Link", Address, lazy=True)
         birthday = f.DateField("Birthday")
+        photo = f.Field("Photo", read_only=True)
```
NicoHood, line 55: Why is this just a "Field" and no "TextField"?

tests/test_orm.py
```diff
@@ -64,20 +65,24 @@ class Meta:
         email="gui@gui.com",
         is_registered=True,
         birthday=datetime(2020, 12, 12).date(),
+        photo="{json: photo}"
```
NicoHood, line 68: what is this line meant to? It seems like invalid json

docs/source/orm.rst
```diff
@@ -23,6 +23,7 @@ Fields
 
 .. automodule:: pyairtable.orm.fields
     :members:
+    :undoc-members:
```
NicoHood, line 26: why was this added?

aleafa: I think it should compile more robust ORM documentation. It's not a necessary addition by any means

tests/test_orm.py
aleafa: It's filler text for a read-only field. Syntax shouldn't matter.

tests/test_orm.py
aleafa: It's simulating a generic Field, not a TextField. Does that result in an unreliable test?

tests/test_orm_fields.py
aleafa: Oh, I think this test is based on a prior iteration of the code. It should be updated.

tests/test_orm.py
NicoHood: as it is confusing, I'd set it to something less confusing like "Test Text" etc.

aleafa, line 68:
```suggestion
        photo="field data"
```

pyairtable/orm/model.py
aleafa: Sure. How about this?

```suggestion
        # Filter fields by read_only attribute so we only set values for writable fields
        map_ = self._field_name_descriptor_map()
```

pyairtable/orm/model.py
```diff
@@ -209,7 +209,8 @@ def save(self) -> bool:
...
+        fields = {f: v for f, v in record["fields"].items() if not _fndm[f].read_only}
```
aleafa, line 213:
```suggestion
        fields = {f: v for f, v in record["fields"].items() if not map_[f].read_only}
```

gtalarico:
Sorry for the delay here @larsakerson will review this and incorporate it

larsakerson:
My apologies. I've incorporated the suggestions; let me know if anything else is needed here, @gtalarico

PR #180, opened 2022-09-19
Reduce unnecessary sleep after API calls

mesozoic:
Right now `ApiAbstract._iterate` calls `time.sleep` before yielding the results of its first API call. This slows down operations that only retrieve a single record (or a single page of results) and it seems unnecessary unless we're about to request a second page of results.

This moves the `time.sleep` call to the end of the loop, so it only happens if we need to fetch a second page.


gtalarico:
thanks @mesozoic

PR #183, opened 2022-09-27
Add time zone support to datetime utilities

larsakerson:
Airtable is [adding more time zone formatting options to datetime fields](https://community.airtable.com/t/new-beta-more-time-zone-formatting-options-for-datetime-fields). This pull request updates the datetime utilities to handle values with time zones other than UTC.


NicoHood:
Why using pytz? Cant we just use `ZoneInfo` which is the way to go with timezones in the future?

aleafa:
@NicoHood honestly I typically use `arrow` for all of this, so I was just looking for a fallback but the change you suggest was easy to make. Thanks for the suggestion.

.gitignore
```diff
@@ -3,6 +3,7 @@
 .vscode
 .venv
 
+.env/
```
NicoHood, line 6: do we need this? .venv seems to be the standard

NicoHood:
what is `arrow`? But the changes look nice, thanks!

aleafa:
[Arrow](https://arrow.readthedocs.io/en/stable) has relieved so many date and time headaches for me.
> Arrow is a Python library that offers a sensible and human-friendly approach to creating, manipulating, formatting and converting dates, times and timestamps. It implements and updates the datetime type, plugging gaps in functionality and providing an intelligent module API that supports many common creation scenarios. Simply put, it helps you work with dates and times with fewer imports and a lot less code.

gtalarico:
hey similar to other prs look there are a few unrelated changes

aleafa:
@gtalarico just learning here... you'll probably want to squash the mess of commits I've made.

Thanks for your patience! Let me know if I need to do anything else.

pyairtable/utils.py
```diff
@@ -21,7 +23,9 @@ def datetime_from_iso_str(value: str) -> datetime:
     Args:
         value: datetime string e.g. "2014-09-05T07:00:00.000Z"
     """
-    return datetime.strptime(value, "%Y-%m-%dT%H:%M:%S.%fZ")
+    if "Z" in value:
```
NicoHood, line 26: Shouldnt we rather check if it and with z? Because from what I understand basically the delimiter could also be a Z, even though that might be never used. But then the parsing would fail anyways, I dont know. Can't we even parse the Z with the %z?

pyairtable/utils.py
```diff
@@ -10,7 +10,9 @@ def datetime_to_iso_str(value: datetime) -> str:
     Args:
         value: datetime object
     """
-    return value.isoformat(timespec="milliseconds") + "Z"
+    if value.utcoffset() is None:
+        return value.isoformat(timespec="milliseconds") + "Z"
```
NicoHood, line 14: Woudln't it make more sense to add a timezone and use the same isoformat function below for all calls? Also is it correct to just add UTC to a datetime object that is not timezone aware? I guess that can end in weird and unwanted behavior.

pyairtable/utils.py
```diff
@@ -10,7 +10,9 @@ def datetime_to_iso_str(value: datetime) -> str:
     Args:
         value: datetime object
     """
-    return value.isoformat(timespec="milliseconds") + "Z"
+    if value.utcoffset() is None:
```
NicoHood, line 13: Dont't we also need to check the `utcoffset`?
https://docs.python.org/3/library/datetime.html#determining-if-an-object-is-aware-or-naive

pyairtable/utils.py
NicoHood: Thinking about this some more time I think we should rather fail here, if not timezone aware object was passed in. Also it seems not correctly implemented in the first place. But I might be wrong, as I've not tested the code, so please feel free to give me some more feedback. Implementing timezones is quite fuzzy, you have to be concentrated.

NicoHood: So I made some tests myself, as I also needed this functionality. You have two options to pass to this function:
a) a timezone aware object
b) a timezone unaware object (naive)

In both cases, the syntax `value.isoformat(timespec="milliseconds")` will work. The `+Z` was not required in first place it seems. Airtable will treat naive datetime strings as UTC.

Personally I'd implement it that naive datetime objects should be added with the local system timezone:
If called without arguments (or with tz=None) the system local timezone is assumed for the target timezone. The .tzinfo attribute of the converted datetime instance will be set to an instance of [timezone](https://docs.python.org/3/library/datetime.html#datetime.timezone) with the zone name and offset obtained from the OS.

pyairtable/utils.py
```diff
@@ -10,7 +10,9 @@ def datetime_to_iso_str(value: datetime) -> str:
...
+        return value.isoformat(timespec="milliseconds") + "Z"
+    return value.isoformat(timespec="milliseconds")
```
NicoHood, line 15:
```suggestion
    # Add local timezone if naiv datetime object was passed in
    # https://docs.python.org/3/library/datetime.html#datetime.datetime.astimezone
    if value.tzinfo is None or value.tzinfo.utcoffset(d) is None:
        value = value.astimezone()

    return value.isoformat(timespec="milliseconds")
```

pyairtable/utils.py
```diff
@@ -21,7 +23,9 @@ def datetime_from_iso_str(value: str) -> datetime:
     Args:
         value: datetime string e.g. "2014-09-05T07:00:00.000Z"
     """
-    return datetime.strptime(value, "%Y-%m-%dT%H:%M:%S.%fZ")
+    if "Z" in value:
+        return datetime.strptime(value, "%Y-%m-%dT%H:%M:%S.%fZ")
+    return datetime.strptime(value, "%Y-%m-%dT%H:%M:%S.%f%z")
```
NicoHood, line 28: For this one the change is also quite simple. The `%z` will already recognize the `Z` from the string. No need to differentiate.
```suggestion
    return datetime.strptime(value, "%Y-%m-%dT%H:%M:%S.%f%z")
```

NicoHood: We could argue if we want the time by airtable to default to our local timezone, as it is by default send as UTC. But with the coming update that might not be the case anymore, so I would not touch that. Since all relevant information is available, and you can always call astimezone, this should be fine now

NicoHood:
I've left a lot of comments. I strongly recommend to accept my suggestions. If there are any questions, please ask. Timezones are a bit fuzzy.

pyairtable/utils.py
aleafa: This looks good. Thanks. Shouldn't the condition evaluate the truth of `value.tzinfo.utcoffset(value) is None`?

pyairtable/utils.py
aleafa: Sounds good. Thank you!

pyairtable/utils.py
aleafa: Thanks for your review! Just one comment and I'd happy to commit the revisions.

pyairtable/utils.py
NicoHood: Ooooh yes. good catch! I've corrected that

pyairtable/utils.py
NicoHood: by the way: why dont we use this?
https://docs.python.org/3/library/datetime.html#datetime.time.fromisoformat

aleafa: That would make sense to me, though it also raises a question for me about how this library parses date fields (i.e., dates without times). Seems like it might not perform as desired

larsakerson:
Thanks, @NicoHood. I've incorporated your suggestions. Let me know if anything else is needed from me here, @gtalarico

mesozoic:
Hi there, I'm going through older PRs in this repo and trying to merge or close them. I see that this PR uses the `zoneinfo` library (which is not present in Python 3.7 or 3.8) and that the tests don't seem to pass:
```
    def test_datetime_utils(datetime_obj, datetime_str):
>       assert utils.datetime_to_iso_str(datetime_obj) == datetime_str
E       AssertionError: assert '2025-12-31T2...:59.555-08:00' == '2025-12-31T23:59:59.555Z'
E         - 2025-12-31T23:59:59.555Z
E         ?                        ^
E         + 2025-12-31T23:59:59.555-08:00
E         ?                        ^^^^^^
```

In my limited experience and testing, the Airtable API will always return timestamps in UTC, regardless of how the base is configured to display them on the web. Given that, can I close this pull request? If there is a use case that I'm missing, can you share some steps to reproduce?

NicoHood:
It may be the case, that the API outputs UTC, but the date that we put into the api from python might not be. Also this hack with hardcoding "Z" as timezone is not very smart. Its just more flexible.

I personally would not care too much about older python versions that are mostly end of life:
https://endoflife.date/python

mesozoic:
> It may be the case, that the API outputs UTC, but the date that we put into the api from python might not be. 

I might've missed the intent here; I thought this was about parsing values from Airtable, not the values we pass in. If the goal is to allow an implementer to pass a timezone-aware `datetime` to the API, I think a patch to that effect (with passing tests showing what new library behavior it offers) would be very welcome. I'm guessing this is for the ORM?

> I personally would not care too much about older python versions that are mostly end of life

This repo has no written compatibility policy (maybe it should) but my instinct is that a widely-used library should keep supporting all live Python versions until _at least_ the official end-of-life. Plenty of organizations still have code running on Python 3.7, and Python 3.8 won't be end-of-life for another year and a half. Maybe we can use [backports.zoneinfo](https://pypi.org/project/backports.zoneinfo/) here.

mesozoic:
Hi, I'm doing another pass through old issues & PRs, and I'm wondering if we should close this one. If there's a legit bug or feature gap for us to address with respect to timezones, I think I could use help reproducing it. Even a code snippet against the Airtable API showing the expected vs actual behavior would be very helpful.

NicoHood:
I think the PR should still get merged. The hardcoded Z might just crash. You can now set specific timezones in airtable. I am not sure if the API will always return UTC, but adding the correct `%z` identifier is just safe and future-proof. A must-have I'd say and a nobrainer.

About the isoformat I would say the `+Z` is also a bug and should not be there. The data should either always have a timezone set or at least be intepreted as a specific timezone. This can be UTC or the local timezone. But always adding +Z is wrong, if the input data has a timezone set. I guess the code will then crash.

I am not sure where those functions are currently being used. I remember only in ORM, but I am not sure. I am not using ORM so I guess I dont have those problems. Timezones are tricky, but we should treat it with care.

I would reopen/merge this. at least line 18 and 29

mesozoic:
@NicoHood I'll be happy to review any future pull requests that follow the contribution guidelines. Right now this branch does not. In either case, without steps to reproduce, it's not clear to me there's a bug.

PR #196, opened 2022-10-20
Fix #194 Create and update commands do not take return_fields_by_field…

### LINKED ISSUE #194: Create and update commands do not take return_fields_by_field_id into account
The record returned will always have names, not IDs. We need to fix it. I guess it is quite simple.

NicoHood:
…_id into account


NicoHood:
@FredZhao-at It seems that this change does not affect anything. Is it possible that the (batch) create and update api command do not support the `returnFieldsByFieldId=1` parameter? Now I am very confused if I am wrong or this is an API bug.

PR #199, opened 2022-10-21
Fix #198 batch update parameter type

### LINKED ISSUE #198: batch_update type unclear
Hi,
I am talking about this code:

```py
    def _batch_update(
        self,
        base_id: str,
        table_name: str,
        records: List[dict],
        replace=False,
        typecast=False,
        **options
    ):
        updated_records = []
        table_url = self.get_table_url(base_id, table_name)
        method = "put" if replace else "patch"
        params = self._options_to_params(**options)
        for records in self._chunk(records, self.MAX_RECORDS_PER_REQUEST):
            chunk_records = [{"id": x["id"], "fields": x["fields"]} for x in records]
            response = self._request(
                method,
                table_url,
                json_data={"records": chunk_records, "typecast": typecast},
                params=params
            )
            updated_records += response["records"]
            time.sleep(self.API_LIMIT)

        return updated_records
```

The type of `records` should be `List[dict]` but it is not clear what the dict should contain. Also I find this confusing, why is this syntax used? Wouldnt it be better to use a type of `dict[str, dict]` where `str` is the id and `dict` the fields? That would be the syntax of the normal update operation. That List feels very unnatural.

NicoHood:
See https://github.com/gtalarico/pyairtable/issues/198

### DIFF
--- a/pyairtable/api/abstract.py
+++ b/pyairtable/api/abstract.py
@@ -76,10 +76,13 @@ class ApiAbstract(metaclass=abc.ABCMeta):
             params.update(to_params_dict(name, value))
         return params
 
-    def _chunk(self, iterable, chunk_size):
+    def _chunk(self, iterable: list|dict, chunk_size):
         """Break iterable into chunks"""
         for i in range(0, len(iterable), chunk_size):
-            yield iterable[i : i + chunk_size]
+            if isinstance(iterable, dict):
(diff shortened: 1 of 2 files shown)


gtalarico:
Thanks for PR but as discussed in other PR "records" will be represented as list of dicts.

Issue #208, opened 2022-11-15
Adding metadata API (2022-11-15 update)


joe-irving:
Airtable has just released a major update to the API, detailed [in the new docs](https://airtable.com/developers/web/api/changelog).

This includes a change to personal access tokens for authentication, and a few new classes (with new methods) which I have tried to lay out below to plan adding them to this wrapper.

Completely new classes and their methods:

* Fields
    * Update Field
* Comments
    * Update comment
    * Delete comment
* Webhooks
     * List payloads
     * Disable/Enable notifications
     * Delete
     * Refresh

And new methods for existing classes (same format):

* Record
    * List Comments
    * Create Comment
* Table
    * List fields
    * List views
    * Create field
    * 'Upsert' option in the batch_update method
* Base
   * Get Schema (List tables)
   * List webhooks
* Api
   * List bases
   
Also added were a few methods for enterprise users only so probably not a priority. 

I don't know if this is the best way to go about tracking adding this functionality, so suggestions welcome.

gtalarico:
This is great. Thanks for organizing this @joe-irving 
Will be tons of work 😮‍💨 😆

BAPCon:
I'm about halfway through webhooks

PR #209, opened 2022-12-04
🚧 Add upsert functionality to batch_update

marks:
## Summary
Addresses 
> 'Upsert' option in the batch_update method

from  #208. Relevant Airtable API docs live at https://airtable.com/developers/web/api/update-multiple-records

## Example usage
Assumes a record already exists with `Unique ID` `aaa`
```python
inputRecords = [
    {'fields': {'Unique ID': 'aaa', 'First Name': 'A'}},
    {'fields': {'Unique ID': 'bbb', 'First Name': 'B'}}
]

upsert_result = table.batch_update(inputRecords, perform_upsert=dict(fieldsToMergeOn=['Unique ID']))
pprint(dict(upsert_result=upsert_result))
```
returns
```
{'upsert_result': {'created_record_ids': ['recIln1E0eUxq0S0i'],
                   'records': [{'createdTime': '2022-10-24T19:49:10.000Z',
                                'fields': {'First Name': 'aaa',
                                           'Name': 'aaa  (ID:aaa)',
                                           'Unique ID': 'aaa'},
                                'id': 'reckUKamPArJEfbwe'},
                               {'createdTime': '2022-12-04T22:12:25.000Z',
                                'fields': {'First Name': 'bbb',
                                           'Name': 'bbb  (ID:bbb)',
                                           'Unique ID': 'bbb'},
                                'id': 'recIln1E0eUxq0S0i'}],
                   'updated_record_ids': ['reckUKamPArJEfbwe']}}
```

### DIFF
--- a/pyairtable/api/abstract.py
+++ b/pyairtable/api/abstract.py
@@ -3,7 +3,7 @@ import posixpath
 import requests
 import time
 from functools import lru_cache
-from typing import List, Optional, Tuple
+from typing import List, Optional, Tuple, TypedDict
 
 from urllib.parse import quote
 from requests.sessions import Session
@@ -14,6 +14,10 @@ from .. import compat
 
 
 TimeoutTuple = Tuple[int, int]
+
+
+class UpsertOptions(TypedDict):
+    fieldsToMergeOn: List[str]
 
 
 class ApiAbstract(metaclass=abc.ABCMeta):
@@ -83,6 +87,16 @@ class ApiAbstract(metaclass=abc.ABCMeta):
 
     def _build_batch_record_objects(self, records):
         return [{"fields": record} for record in records]
+
+    def _build_upsert_batch_record_list(self, records):
+        chunk_records = []
+        for x in records:
+            new_dict = dict(fields=x["fields"])
+            id = x.get("id", None)
+            if id:
+                new_dict["id"] = id
+            chunk_records.append(new_dict)
+        return chunk_records
 
     def _process_response(self, response):
         try:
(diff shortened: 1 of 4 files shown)


marks:
Hiya @gtalarico - first, thank you for this great library. This PR is very much a WIP but wanted to get your thoughts on it. A few things are top of mind for me:
1. [The introduction of `TypedDict`](https://github.com/gtalarico/pyairtable/pull/209/files#diff-8af5214420de7df4c6dd9f657201ccf78f6f8891391040d4880f5af26261bcffR19-R20)
2. Adding this functionality to the existing batch_update methods -- since there is a different return format, should this be a whole new function? (IMO, that would be a better end-user experience)
3. Or, to keep it simple, just return an array of records (this would mean not exposing information to the user which the API does return - IDs of created and updated records respectively)

gtalarico:
Thanks Mark sorry I haven't had a chance to review this. On my to-do list

MarekVigas:
Hey, can I help with this PR? What needs to be done to get this merged and released? I want to use it in my scripts.

mesozoic:
Hi there! I'm trying to go through some older PRs in this repo and get them across the line. I'd love to see upsert functionality get merged in too, so I'm happy to help review this PR.

Personally, I think this makes most sense as its own method, since the mechanics are different, the return type is different (extra fields), and there is at least one required parameter (fields to merge) that isn't required for an update. Separating them out creates less opportunity to confuse the two behaviors or introduce bugs between the two. 

Perhaps something with a signature like this?

```python
class ApiAbstract:
    def _upsert(
        self,
        base_id: str,
        table_name: str,
        records: Sequence[Dict],
        merge_fields: Sequence[str],
        replace: bool = False,
        typecast: bool = False,
        return_fields_by_field_id: bool = False,
    ) -> List[Dict]:
        ...
```

I think we will also need to write a few tests for this endpoint, given how many different conditional behaviors they describe in the API documentation. 

For what it's worth, I wouldn't worry about introducing TypedDict right now. I'm working on a branch to implement strict typing across the entire library and it's moderately disruptive, so we're better off doing it all in one merge vs. trying to introduce it piecemeal with each new feature.

Issue #211, opened 2022-12-12
Early heads up: switching away from user API keys


FredZhao-at:
Hi! I’m Fred, an engineer on the Airtable API team. I’m writing here to share some plans we have to move away from user API keys, with a goal of communicating it early so you have time to triage and prioritize work needed.

First, some context. As #208 mentioned, we recently published a new developer doc site, and announced two new authentication methods, as well as new endpoints and capabilities those methods support: https://airtable.com/developers/web/api/changelog#anchor-2022-11-15 

Since these new authentication methods (personal access tokens and OAuth integration access tokens) are much more secure than the current user API key authentication method, we are limiting the new endpoints and capabilities to only be available to the new methods.

The medium term plan is to deprecate user API keys. We are still working on the exact details, but the deprecation period will last 1 year, and we expect it to begin at the start of next year (Jan 2023).

Based on this, we recommend preparing to support the new API token format:
* If you currently validate tokens (e.g. with a regex like `/^key[a-zA-Z0-9]{14}`), start supporting the new personal access token key format as well (`/^pat[a-zA-Z0-9]{14}\.[0-9a-f]{64}`)
    * **For pyairtable**: From a quick scan, I don't think this applies, so I'm mentioning this point more for completeness
* Update documentation mentions of “API key” to the more general “API key or access token”
    * **For pyairtable**: This may apply to the article at https://pyairtable.readthedocs.io/en/latest/getting-started.html#api-key

gtalarico:
Thanks for the details @FredZhao-at 
I am pretty low on bandwidth at the this time. Happy to take PRs for this, otherwise I will tackle it on when time allows.

We can add the regex validations, and a deprecation warning for the soon old api key format.

FredZhao-at:
> We can add the regex validations, and a deprecation warning for the soon old api key format.

To clarify, regex validation is not actually necessary. The fact that you're treating it as an opaque string right now is actually great, since it means less work needed. :) Sorry for the noise in that bullet point, I'll edit it to say "no action needed"

gtalarico:
Docs were taken care on #214 
We could add regex check just for the purpose of emitting deprecation warning if you all think it's helpful

mesozoic:
Looks like this was taken care of and no further action needed (yay), so I'm going to close the issue.

PR #214, opened 2022-12-16
Changes "API Key" to "API Key or Access Token"

BAPCon:
Due to the slow deprecation of API keys, it was mentioned in issue #211 that the docs should reflect access token usage.
If different verbiage is desired, please let me know!

### DIFF
--- a/docs/source/api.rst
+++ b/docs/source/api.rst
@@ -21,13 +21,13 @@ For example, the three ``all()`` calls below would return the same result:
 
   from pyairtable import Api, Base, Table
 
-  api = Api('apikey/accesstoken')
+  api = Api('auth_token')
   api.all('base_id', 'table_name')
 
-  base = Base('apikey/accesstoken', 'base_id')
+  base = Base('auth_token', 'base_id')
   base.all('table_name')
 
-  table = Table('apikey/accesstoken', 'base_id', 'table_name')
+  table = Table('auth_token', 'base_id', 'table_name')
   table.all()
 
 Interface
@@ -215,7 +215,7 @@ specified on each request.
 
 .. code-block:: python
 
-  >>> base = Base('apikey/accesstoken', 'base_id')
+  >>> base = Base('auth_token', 'base_id')
   >>> base.all('Contacts)
   [{id:'rec123asa23', fields': {'Last Name': 'Alfred', "Age": 84}, ... ]
 
@@ -265,7 +265,7 @@ Default Retry Strategy
 .. code-block:: python
 
   from pyairtable import Api, retry_strategy
-  api = Api('apikey/accesstoken', retry_strategy=retry_strategy())
+  api = Api('auth_token', retry_strategy=retry_strategy())
 
 
 Adjusted Default Strategy
@@ -273,7 +273,7 @@ Adjusted Default Strategy
 .. code-block:: python
 
   from pyairtable import Api, retry_strategy
-  api = Api('apikey/accesstoken', retry_strategy=retry_strategy(total=3))
+  api = Api('auth_token', retry_strategy=retry_strategy(total=3))
 
 Custom Retry
 
@@ -283,7 +283,7 @@ Custom Retry
   from urllib3.util import Retry
 
   myRetry = Retry(**kwargs)
-  api = Api('apikey/accesstoken', retry_strategy=myRetry)
+  api = Api('auth_token', retry_strategy=myRetry)
 
 
 .. autofunction:: pyairtable.api.retrying.retry_strategy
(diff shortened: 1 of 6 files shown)


NicoHood:
But there is no such parameter `accesstoken`!?

NicoHood:
But the python parameter name does not exist. We cannot document a parameter, that does not exist in the python api. We should rather document in a comment/text/description of that parameter, that it also accepts access token. but we cannot rename it to something that will not work

marks:
> But the python parameter name does not exist. We cannot document a parameter, that does not exist in the python api. We should rather document in a comment/text/description of that parameter, that it also accepts access token. but we cannot rename it to something that will not work

I could have missed something, but it seemed like all of the proposed changes in this PR were in documentation and comments, not code

NicoHood:
Yeah, but they suggest that the syntax is a valid parameter

NicoHood:
Alright, so I see where we might missunderstand.

I usually use kwargs, meaning I do `Api(api_key='api_key')` and not `Api('api_key')`. To me this looks like the name inside the string (okay it is a string and not the keyword) is the actual parameter.

I would go even further and change the docs and specify the exact keyword/parameter. Maybe then it also does not matter too much what the content of the string itself is.

Does this sound like a solution?

docs/source/getting-started.rst
```diff
@@ -19,10 +19,10 @@ Installation
 _______________________________________________
 
 
-Api Key
+Api Key / Access Token
```
gtalarico, line 22: Authorization Token

docs/source/getting-started.rst
```diff
@@ -19,10 +19,10 @@ Installation
...
 *******
 
-Your Airtable API key should be securely stored. 
+Your Airtable API key or access token should be securely stored.
```
gtalarico, line 25: > Airbnb accepts two types of authentication tokens: Api Keys, and Personal Access tokens. 
> Your auth token should be securely stored.

docs/source/substitutions.rst
```diff
@@ -1,6 +1,6 @@
 
 
-.. |arg_api_key| replace:: An Airtable API Key.
+.. |arg_api_key| replace:: An Airtable API Key or Access Token.
```
gtalarico, line 3: An Airtable Authorization Token

gtalarico:
Given all tokens are added to `Authorization: Bearer <token>` I think we can just call it `auth_token` in code (arg name in code and doc string).

And then in the docs (examples, getting started, etc)  we can clarify that auth token can be either a legacy "Api Key" or **Access Token**.

As far as I understand, "Access Token" include both personal access tokens created here via https://airtable.com/create/tokens, as well as tokens returned by [token creation request](https://airtable.com/developers/web/api/oauth-reference#token-creation-request)

This will also make it easier to deal with when the legacy api keys are deprecated (only update doc).

Leaving some comments in PR with my suggestions

pyairtable/api/api.py
```diff
@@ -8,14 +8,14 @@ class Api(ApiAbstract):
     """
     Represents an Airtable Api.
 
-    The Api Key is provided on init and ``base_id`` and ``table_id``
+    The Api Key or Access Token is provided on init and ``base_id`` and ``table_id``
```
gtalarico, line 11: Authorization Token is provided...

docs/source/api.rst
```diff
@@ -21,13 +21,13 @@ For example, the three ``all()`` calls below would return the same result:
 
   from pyairtable import Api, Base, Table
 
-  api = Api('apikey')
+  api = Api('apikey/accesstoken')
```
gtalarico, line 24: `Api('auth_token')`

docs/source/api.rst
```diff
@@ -21,13 +21,13 @@ For example, the three ``all()`` calls below would return the same result:
...
   api.all('base_id', 'table_name')
 
-  base = Base('apikey', 'base_id')
+  base = Base('apikey/accesstoken', 'base_id')
```
gtalarico, line 27: ditto

pyairtable/api/api.py
```diff
@@ -8,14 +8,14 @@ class Api(ApiAbstract):
...
     can be provided on each method call.
 
     If you are only operating on one Table, or one Base, consider using
     :class:`Base` or :class:`Table`.
 
     Usage:
-        >>> api = Api('apikey')
+        >>> api = Api('apikey/accesstoken')
```
gtalarico, line 18: ditto

docs/source/getting-started.rst
gtalarico: and below

```
.. Note: 
     Personal access tokens are a new, more secure way to grant API access to your Airtable data.
     They can create multiple tokens, grant them access to specific bases, and manage them individually.
 
     < Some note about how Api Keys are scheduled to deprecated>
```

gtalarico: @marks

gtalarico (commented):
Thanks for taking a stab at this

docs/source/api.rst
BAPCon: Sounds good

docs/source/getting-started.rst
BAPCon: So is the header there just "Authorization Token" or "Api Key / Authorization Token"

PR #217, opened 2023-01-11
Run mypy as part of `make lint`, `tox -e lint`

mesozoic:
I noticed that mypy isn't run or enforced as any part of the build chain, so I added it to the `lint` targets. This might make it easier to catch any issues in pull requests in the future. (I've got a couple I'd like to send in soon.)

I couldn't really tell what the `pyairtable.compat` module was for, but mypy didn't seem to like it ([error](https://i.fluffy.cc/SJMKhjPbSbd5JMsFXng3vSwVcmm1L7fn.html)) and removing it didn't break any of the tests.


gtalarico:
It's been a while but I think the `compat` was added as a place to stash  logic to handle typing when optional imports are used `Retry`.
https://stackoverflow.com/a/61394121/4411196
I do see the error you mentioned, we need to fix it.

mesozoic:
That's a good catch, thanks! I bumped up to the earliest version of requests that has a vendored library, which is still a few years old (so presumably few people out there still need older versions). To ensure nobody else makes the same mistake I did, I've added an extra environment specifier to `tox.ini` that will run tests on both the minimum required version _and_ the latest version of the requests library.

It looks like #182 added a couple other mypy failures ([saved here](https://i.fluffy.cc/bxhwprVzlxfPCFHSxKLbvknk03sRrG42.html)), and I wasn't quite sure what to make of them. I did my best to sort it out, and it looks like there was some unused code that could just be removed. Open to feedback if there's something I missed there, however. @larsakerson FYI

mesozoic:
Actually, reading https://github.com/python/typeshed/issues/6893, it looks like we can't rely on the vendored version of the urllib3 library once we use the latest version of mypy and type stubs. So I've switched the code to refer directly to urllib3, and now the tests pass even farther back (requests==2.8.0). I believe this is ready to merge but still open to things I've missed.

mesozoic:
I can see the docs build has failed. Bear with me, I'm learning here... will push again once I've figured it out.

mesozoic:
Looking at [this build failure](https://readthedocs.org/projects/pyairtable/builds/20051009/) readthedocs.org is still building on Python 3.7, which means we can't use `flake8 >= 6.0.0`. I've removed the minimum version requirements from linting dependencies. I've also moved the docs build into tox so it's more reproducible.

It looks like the version of Python used for building documentation [can be configured](https://docs.readthedocs.io/en/stable/config-file/v2.html) but I'm not trying to take that on right now. I did encounter one formatter warning on the existing documentation, which I've fixed.

gtalarico:
Good with me

PR #218, opened 2023-01-17
Sleep Time Optimization

alpancs:
sleep times can be optimized (reduced) by calculating elapsed times between iterations

### DIFF
--- a/pyairtable/__init__.py
+++ b/pyairtable/__init__.py
@@ -1,4 +1,4 @@
-__version__ = "1.4.1"
+__version__ = "1.4.0"
 
 from .api import Api, Base, Table  # noqa
 from .api.retrying import retry_strategy  # noqa
--- a/pyairtable/__init__.py
+++ b/pyairtable/__init__.py
@@ -1,4 +1,4 @@
-__version__ = "1.4.0"
+__version__ = "1.4.1"
 
 from .api import Api, Base, Table  # noqa
 from .api.retrying import retry_strategy  # noqa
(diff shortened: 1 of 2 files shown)


pyairtable/__init__.py
```diff
@@ -1,4 +1,4 @@
-__version__ = "1.4.0"
```
NicoHood, line 1: I guess that does not belong to this PR/feature

alpancs: will revert then

alpancs:
All checks have passed.
Please review once again @NicoHood.
Thanks!

alpancs:
please also review @gtalarico
thank you

gtalarico:
This looks good, thanks!

 Have you tried this? Curious what the actual performance difference would be.
Could you do a few timeit against both?

gtalarico:
Happy to merge this but would like to see a comparison of performance w/. and w/out this.
Want to make sure that the overhead of fetching time on each loop is worth

mesozoic:
Just offering my two cents, instead of adding complexity to the sleep behavior, I'd love to see this library aim to remove `time.sleep` calls entirely. It could instead rely on a default retrying behavior that accomplishes the same thing. 

The reason is that it's very possible to have multiple processes hitting the API simultaneously with no awareness of each other, and in that circumstance there's no way to predict the appropriate amount of time to wait between API calls. Better to let the Airtable API throttle requests and insert escalating amounts of delay when you get 429s. That's what their official JS library does ([code](https://github.com/Airtable/airtable.js/blob/2d36eb64249a51e99cc7f2ea3dfb59cd3522a188/src/run_action.ts#L70)).

gtalarico:
@mesozoic agree 100%

alpancs:
sorry @gtalarico, I haven't had time to manage this PR, but I agree with @mesozoic. I also had the same concern about the multi-process usage from the beginning.

should this PR be modified to remove `time.sleep` entirely?

PR #220, opened 2023-01-18
Update license metadata in setup.cfg

mesozoic:
Some automated scanning tools rely on consistent identifiers to determine what license applies to a library. "MIT" is the most common spelling of the MIT License used [(see packaging.python.org)](https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#license) and the OSI classifier is unambiguous. This branch adds both of those to setup.cfg.


gtalarico:
Thank you 🙏

PR #226, opened 2023-02-12
Fix typos in docstring, docs

meli-lewis:
Fairly straightforward, though because I have limited experience with Sphinx, I'm not sure whether editing the .rst was sufficient to update the docs entirely.


mesozoic:
Thanks for catching these!

Issue #230, opened 2023-02-26
Add pagination support for (metadata) get base list endpoint


marks:
Airtable API docs at https://airtable.com/developers/web/api/list-bases note that 1,000 bases are returned per request. You can paginate with an `offset` param

https://github.com/gtalarico/pyairtable/blob/0f9fd8e7c320b065761033a32d7f0eba0abda3ed/pyairtable/metadata.py#L6-L32 in this library should get updated accordingly

PR #231, opened 2023-02-27
Update get_api_bases to paginate with offset cursor

marks:
Currently, if the token used to call the [list bases endpoint](https://airtable.com/developers/web/api/list-bases) has access to more than 1,000 bases, pyairtable will return only the first page of 1,000 bases.

This PR fixes that.

Issue #230

### DIFF
--- a/docs/source/metadata.rst
+++ b/docs/source/metadata.rst
@@ -1,21 +1,16 @@
 .. include:: substitutions.rst
 
-Metadata Api
-==============
+Metadata
+========
 
 .. versionadded:: 1.0.0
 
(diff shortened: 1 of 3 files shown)


mesozoic:
I will admit I've never had more than a thousand bases so I've never run into this 😁  

Are you able to add a test that exercises the behavior in question? Right now the test only exercises the null behavior (when no `offset` parameter is present).

mesozoic:
This was fixed in 75203c6d3fd12bd90fc3d3497567f88de139f3d1, part of #282

PR #236, opened 2023-03-29
Add webhook methods

BAPCon:
Adds webhook methods to abstract and to base, as referenced in #208 
I think separating everything out into a separate meta.py file might be better.

### DIFF
--- a/pyairtable/api/abstract.py
+++ b/pyairtable/api/abstract.py
@@ -238,6 +238,42 @@ class ApiAbstract(metaclass=abc.ABCMeta):
             time.sleep(self.API_LIMIT)
         return deleted_records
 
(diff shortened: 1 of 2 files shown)


BAPCon:
Accidentally deleted a previous pull request.
Let me know if any changes are needed, I don't mind doing it myself.
I am new to the process of writing documentation but would like to learn.

mesozoic:
I agree with separating out metadata from the ApiAbstract class into its own module, aside from maybe a single helper method for constructing a URL and calling an endpoint w/ the appropriate headers. The metadata APIs have quite a bit to them, and it's arguably clutter to put them all in the same namespace where developers are interacting with the contents of the base itself.

PR #238, opened 2023-03-29
Add Various Comment Methods to Table class

BAPCon:
Adds methods for listing comments in a record, updating comments, creating new comments, and deleting existing comments.
Part of a number of Airtable API updates as referenced in #208

### DIFF
--- a/pyairtable/api/table.py
+++ b/pyairtable/api/table.py
@@ -161,6 +161,9 @@ class Table(ApiAbstract):
             offset = meta_response.get('offset')
         return comments
     
+    def get_comments_url(self, record_id: str, comment_id: str = ''):
+        return super()._get_comments_url(self.base_id, self.table_name, record_id, comment_id)
+    
     def update_comment(self, record_id: str, comment_id: str, text: str):
         return super()._update_comment(self.base_id, self.table_name, record_id, comment_id, text)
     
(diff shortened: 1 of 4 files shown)


gtalarico:
Hey @BAPCon 
Thanks for submission - this looks great.
Would you be able to add some unit tests to go along with these changes?
Thanks

BAPCon:
@gtalarico 
Yeah no problem

BAPCon:
@gtalarico I'm new to writing tests, does this look correct to you?

pyairtable/api/table.py
```diff
@@ -150,6 +150,28 @@ def batch_delete(self, record_ids: List[str]):
...
+        meta_response = super()._get_comments(self.base_id, self.table_name, record_id)
+        comments = meta_response.get('comments')
+        offset = meta_response.get('offset')
+        while offset is not None:
+            meta_response = super()._get_comments(self.base_id, self.table_name, record_id, offset=offset)
+            comments += meta_response.get('comments')
+            offset = meta_response.get('offset')
+        return comments
+    
+    def get_comments_url(self, record_id: str, comment_id: str = ''):
```
gtalarico, line 164: why is comment_id optional? Is method intended to be used for both single fetch + list fetch?

BAPCon: I originally just added that in for the tests, but I've gone and removed it now. There is no single get comment method, it was referring to getting the URL for comments, as the request methods get/patch/post/delete are used rather than separate resources.
The url's are built in each method for getting comments, rather than a separate method.

PR #239, opened 2023-04-10
Use pre-commit and tox, then apply black/flake8 to all files

mesozoic:
[pre-commit](pre-commit.com) is a great framework for applying consistent rules across a repository on every commit, in a way that's not very disruptive for contributors. When coupled with highly opinionated formatters (like [black](https://github.com/psf/black)), it keeps diffs small and avoids any argument over stylistic disagreements.

In order to make these hooks' behavior as reproducible as possible, this branch moves linting/formatting commands into `tox.ini`. This branch also applies the black, flake8, and isort formatters across all files so that contributors have a reasonable baseline to work against. Reviewing one commit at a time might be easier than reviewing all at once.

### DIFF
--- a/pyairtable/api/abstract.py
+++ b/pyairtable/api/abstract.py
@@ -32,7 +32,6 @@ class ApiAbstract(metaclass=abc.ABCMeta):
         timeout: Optional[TimeoutTuple] = None,
         retry_strategy: Optional[Retry] = None,
     ):
-
         if not retry_strategy:
             self.session = Session()
         else:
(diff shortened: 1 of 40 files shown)


pyairtable/api/abstract.py
```diff
@@ -146,21 +143,26 @@ def _all(self, base_id: str, table_name: str, **options) -> List[dict]:
             all_records.extend(records)
         return all_records
 
-    def _create(self, base_id: str, table_name: str, fields: dict, typecast=False, **options):
-
+    def _create(
```
gtalarico, line 146: are we changing line length?

gtalarico (approved):
LGTM

pyairtable/api/abstract.py
mesozoic: No changes. I think this line was already over 88 chars, which is what we have as [our existing line limit today](https://github.com/gtalarico/pyairtable/blob/e367ae9880ff07e789ab8a462dc719ed9962cc2a/pyproject.toml#L7).

gtalarico: Ah weird it thought we were linting on format and preventing merge but I guess that may have stopped working at some point

PR #240, opened 2023-04-10
Run GitHub Actions on every pull request

mesozoic:
Right now GitHub Actions only trigger on a push. This will trigger them on every pull request against `main`, which should have the added benefit of putting them into the "Checks" UI at the bottom of the pull request.


gtalarico (approved):
LGTM

PR #243, opened 2023-04-17
configurable api_base_url

goksan:
Adds configurable `api_base_url` when creating `Api`

```python
api = Api(token, api_base_url="https://api.airwalker.io/")
```

airtable.js implements this as `endpointUrl` 
https://github.com/Airtable/airtable.js/blob/2d36eb64249a51e99cc7f2ea3dfb59cd3522a188/README.md?plain=1#L68

### DIFF
--- a/pyairtable/api/abstract.py
+++ b/pyairtable/api/abstract.py
@@ -21,7 +21,7 @@ class ApiAbstract(metaclass=abc.ABCMeta):
     DEFAULT_API_BASE_URL = "https://api.airtable.com/"
 
     session: Session
-    api_base_url = DEFAULT_API_BASE_URL
+    endpoint_url = DEFAULT_API_BASE_URL
     api_url: str
     tiemout: TimeoutTuple
 
@@ -30,17 +30,17 @@ class ApiAbstract(metaclass=abc.ABCMeta):
         api_key: str,
         timeout: Optional[TimeoutTuple] = None,
         retry_strategy: Optional[Retry] = None,
-        api_base_url: Optional[str] = None,
+        endpoint_url: Optional[str] = None,
     ):
         if not retry_strategy:
             self.session = Session()
         else:
             self.session = _RetryingSession(retry_strategy)
 
-        if api_base_url:
-            self.api_base_url = api_base_url
+        if endpoint_url:
+            self.endpoint_url = endpoint_url
 
-        self.api_url = posixpath.join(self.api_base_url, self.VERSION)    
+        self.api_url = posixpath.join(self.endpoint_url, self.VERSION)    
         self.timeout = timeout
         self.api_key = api_key
 
--- a/pyairtable/api/abstract.py
+++ b/pyairtable/api/abstract.py
@@ -16,12 +16,13 @@ TimeoutTuple = Tuple[int, int]
 
 class ApiAbstract(metaclass=abc.ABCMeta):
     VERSION = "v0"
-    API_BASE_URL = "https://api.airtable.com/"
     API_LIMIT = 1.0 / 5  # 5 per second
-    API_URL = posixpath.join(API_BASE_URL, VERSION)
     MAX_RECORDS_PER_REQUEST = 10
+    DEFAULT_API_BASE_URL = "https://api.airtable.com/"
 
     session: Session
+    api_base_url = DEFAULT_API_BASE_URL
+    api_url: str
     tiemout: TimeoutTuple
 
     def __init__(
@@ -29,12 +30,17 @@ class ApiAbstract(metaclass=abc.ABCMeta):
         api_key: str,
         timeout: Optional[TimeoutTuple] = None,
         retry_strategy: Optional[Retry] = None,
+        api_base_url: Optional[str] = None,
     ):
         if not retry_strategy:
             self.session = Session()
         else:
             self.session = _RetryingSession(retry_strategy)
 
+        if api_base_url:
+            self.api_base_url = api_base_url
+
+        self.api_url = posixpath.join(self.api_base_url, self.VERSION)    
         self.timeout = timeout
         self.api_key = api_key
 
@@ -55,7 +61,7 @@ class ApiAbstract(metaclass=abc.ABCMeta):
     @lru_cache()
     def get_table_url(self, base_id: str, table_name: str):
         url_safe_table_name = quote(table_name, safe="")
-        table_url = posixpath.join(self.API_URL, base_id, url_safe_table_name)
+        table_url = posixpath.join(self.api_url, base_id, url_safe_table_name)
         return table_url
 
     @lru_cache()
--- a/pyairtable/api/abstract.py
+++ b/pyairtable/api/abstract.py
@@ -21,7 +21,6 @@ class ApiAbstract(metaclass=abc.ABCMeta):
 
     session: Session
     endpoint_url: str
-    api_url: str
     tiemout: TimeoutTuple
 
     def __init__(
@@ -36,8 +35,11 @@ class ApiAbstract(metaclass=abc.ABCMeta):
         else:
             self.session = _RetryingSession(retry_strategy)
 
-        self.endpoint_url = endpoint_url
-        self.api_url = posixpath.join(self.endpoint_url, self.VERSION)    
+        if not endpoint_url.endswith(self.VERSION):
+            self.endpoint_url = posixpath.join(endpoint_url, self.VERSION)
+        else:
+            self.endpoint_url = endpoint_url
+        print(self.endpoint_url)
         self.timeout = timeout
         self.api_key = api_key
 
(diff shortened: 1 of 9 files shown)


pyairtable/api/abstract.py
```diff
@@ -16,25 +16,31 @@
...
+    api_base_url = DEFAULT_API_BASE_URL
+    api_url: str
     tiemout: TimeoutTuple
 
     def __init__(
         self,
         api_key: str,
         timeout: Optional[TimeoutTuple] = None,
         retry_strategy: Optional[Retry] = None,
+        api_base_url: Optional[str] = None,
```
mesozoic, line 33: I actually kinda like the name `endpoint_url` (borrowing from the official Airtable library) since "Api" and "Base" are both overloaded terms in this context. Would you object to renaming it?

pyairtable/api/abstract.py
```diff
@@ -16,25 +16,31 @@
 
 class ApiAbstract(metaclass=abc.ABCMeta):
     VERSION = "v0"
-    API_BASE_URL = "https://api.airtable.com/"
     API_LIMIT = 1.0 / 5  # 5 per second
-    API_URL = posixpath.join(API_BASE_URL, VERSION)
     MAX_RECORDS_PER_REQUEST = 10
+    DEFAULT_API_BASE_URL = "https://api.airtable.com/"
```
mesozoic, line 21: We probably don't need this as a class attribute at all, do we? It could just be the default param in `__init__`

mesozoic (changes requested):
Thanks for the patch! Can we add a test (perhaps to `test_api_api.py`) to ensure that passing the parameter correctly changes the URLs we hit?

pyairtable/api/abstract.py
goksan: No objections, what do you think about the naming of `api_url`?

goksan: https://github.com/gtalarico/pyairtable/pull/243/commits/d1f732524461e27d19a350805bf415f1f116516d

pyairtable/api/abstract.py
goksan: https://github.com/gtalarico/pyairtable/pull/243/commits/18a4070dfbbf880f9dcf004bac3b70be153035a9

https://github.com/gtalarico/pyairtable/pull/243/commits/5068bca7090758e967ccad7f617bee75c3668378

goksan:
Added a few tests

https://github.com/gtalarico/pyairtable/blob/0799ba6ca7e9cc0af530e5b92083bd534716b999/tests/test_api_api.py#L23

tests/test_api_api.py
```diff
@@ -19,6 +19,18 @@ def test_get_table(api: Api):
...
+    assert api.endpoint_url == "https://api.airtable.com"
+
+
+def test_endpoint_url(api_with_endpoint_url: Api):
+    assert api_with_endpoint_url.endpoint_url == "https://api.example.com"
+
+
+def test_api_url(api_with_endpoint_url_trailing_slash: Api):
+    assert api_with_endpoint_url_trailing_slash.api_url == "https://api.example.com/" + api_with_endpoint_url_trailing_slash.VERSION
+
```
goksan, line 33: Added a few tests

pyairtable/api/abstract.py
```diff
@@ -16,25 +16,28 @@
...
+    endpoint_url: str
+    api_url: str
     tiemout: TimeoutTuple
 
     def __init__(
         self,
         api_key: str,
         timeout: Optional[TimeoutTuple] = None,
         retry_strategy: Optional[Retry] = None,
+        endpoint_url: Optional[str] = "https://api.airtable.com",
```
mesozoic, line 32: Since `None` is not a valid value, I think this should be `str` instead of `Optional[str]`.

pyairtable/api/abstract.py
```diff
@@ -16,25 +16,28 @@
...
     ):
         if not retry_strategy:
             self.session = Session()
         else:
             self.session = _RetryingSession(retry_strategy)
 
+        self.endpoint_url = endpoint_url
+        self.api_url = posixpath.join(self.endpoint_url, self.VERSION)
```
mesozoic, line 40: I might just combine these two; will we ever need the endpoint _without_ the version number? Would we be just as well off removing one of them and just storing the combined URL?

tests/conftest.py
```diff
@@ -36,6 +36,16 @@ def api(constants):
...
 
 
+@pytest.fixture()
+def api_with_endpoint_url(constants):
+    return Api(constants["API_KEY"], endpoint_url="https://api.example.com")
+
+
+@pytest.fixture()
+def api_with_endpoint_url_trailing_slash(constants):
+    return Api(constants["API_KEY"], endpoint_url="https://api.example.com/")
```
mesozoic, line 46: This probably doesn't need to be a fixture; it could just be a one-line test. But it's a good edge case to test!

mesozoic (changes requested):
Thanks for the edits! A couple minor things to tweak and then I'll be happy to merge this.

pyairtable/api/abstract.py
goksan: https://github.com/gtalarico/pyairtable/pull/243/commits/f1937d7b73e8c0f7cd339adffb0f58bfea8b33a7

pyairtable/api/abstract.py
goksan: https://github.com/gtalarico/pyairtable/pull/243/commits/8eeaa44204ac2d058676b8166fcdd8de1ffa8fb2

tests/conftest.py
goksan: https://github.com/gtalarico/pyairtable/pull/243/commits/5ad5c8012a55d0eef06098f684cc281e64201835

Let me know whether or not I misunderstood your suggestion

pyairtable/api/abstract.py
```diff
@@ -16,25 +16,30 @@
...
         if not retry_strategy:
             self.session = Session()
         else:
             self.session = _RetryingSession(retry_strategy)
 
+        if not endpoint_url.endswith(self.VERSION):
+            self.endpoint_url = posixpath.join(endpoint_url, self.VERSION)
+        else:
+            self.endpoint_url = endpoint_url
+        print(self.endpoint_url)
```
mesozoic, line 42: Please remove the `print()` – I'm guessing that was probably just leftover from debugging? Otherwise lgtm.

(FWIW, I don't think we need to support two different URL formats, it's probably fine to just use the provided value as-is.)

mesozoic: Oh, one more request: can you add a mention of this new parameter in the API documentation?

goksan: > Please remove the `print()` – I'm guessing that was probably just leftover from debugging? Otherwise lgtm.

Whoops https://github.com/gtalarico/pyairtable/pull/243/commits/66bc47941aaa1ed533ed373f2fbebef28e4297af

> Oh, one more request: can you add a mention of this new parameter in the API documentation?

Copied and adjusted from airtable.js
https://github.com/gtalarico/pyairtable/pull/243/commits/14e5d49fd8682b0db099122c611099b299368e06

mesozoic:
Thanks for submitting this patch and working through revisions! ❤️  Due to commit signing requirements on this repository, I've squashed and merged this branch into main as 0189d9d.

goksan:
@mesozoic 

There is a regression between our versions caused by the `endpoint_url` assignment difference.

Here's a snippet that produces an error on main (a valid token, base id and table id is used, I just removed them to share this here):
```python
from pyairtable import Api
token = "TOKEN"
api = Api(token, endpoint_url="https://api.airwalker.io")
table = api.get_table('BASEID', 'TABLEID').create({"Task": "aaa"})
```

```
line 1021, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: ('405 Client Error: Method Not Allowed for url: https://api.airwalker.io/v0/v0/BASEID/TABLEID', '405 Client Error: Method Not Allowed for url: https://api.airwalker.io/v0/v0/BASEID/TABLEID')
```

pyairtable/api/abstract.py
```diff
@@ -16,25 +16,30 @@
...
     ):
         if not retry_strategy:
             self.session = Session()
         else:
             self.session = _RetryingSession(retry_strategy)
 
+        if not endpoint_url.endswith(self.VERSION):
+            self.endpoint_url = posixpath.join(endpoint_url, self.VERSION)
+        else:
+            self.endpoint_url = endpoint_url
```
goksan, line 41: To be clear it's this part

mesozoic: Ah, thanks for pointing out my mistake. This seemed superfluous at first but now I understand it was how you were avoiding duplicating the `/v0` when passing endpoint_url from Api to Table. We don't have a test for this behavior, so I'll add one and fix the issue.

PR #245, opened 2023-04-22
Use `POST /listRecords` instead of `GET` when the URL gets too long

mesozoic:
This attempts to resolve the same issue as #222 but without as much refactoring. It was hard for me to wrap my head around everything going on in that branch, so I started by writing a unit test and an integration test and then wrote only what I needed to get them to pass. I'm open to feedback or suggestions, especially if anything else in that branch is a must-have.

### DIFF
--- a/pyairtable/api/abstract.py
+++ b/pyairtable/api/abstract.py
@@ -2,7 +2,7 @@ import abc
 import posixpath
 import time
 from functools import lru_cache
-from typing import List, Optional, Tuple
+from typing import Dict, List, Optional, Tuple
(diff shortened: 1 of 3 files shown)


mesozoic:
@gtalarico following our discussion, I'm going to merge bugfixes and features after about a week of being up for review if there's no open feedback. Thanks again for your support!

mesozoic:
I reverted the merge; this branch does not appropriately handle all query parameters, which is something I missed porting over from the other PRs that have tried to tackle this. I'll take another pass at it after writing some more comprehensive tests.

PR #247, opened 2023-04-30
Use `POST /listRecords` instead of `GET` when the URL gets too long

mesozoic:
I tried this in #245 but failed to consider enough of the differences between the formatting of query parameters vs. POST parameters. (I reverted that merge in 090bf68.) This branch adds more tests to ensure that the API accepts each parameter we might generate. 

I'll leave this open for a week or so in case there's any feedback, especially if I've missed tests for any other edge cases.

### DIFF
--- a/docs/source/api.rst
+++ b/docs/source/api.rst
@@ -303,8 +303,8 @@ have a corresponding ``kwargs`` that can be used with fetching methods like :met
    :widths: 25 25 50
    :header-rows: 1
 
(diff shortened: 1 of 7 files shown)


mesozoic:
I have a couple other branches that depend on this one, so I'll plan to merge it tomorrow unless there are any objections.

gtalarico (approved):
LGTM

PR #250, opened 2023-05-03
Ignore extra fields in ORM (fixes #190, option 1)

### LINKED ISSUE #190: ORM make fields optional
Hi,
as you can see I am playing around with ORM.

I have a table that has lots of attributes, some of them are computed fields. I know we have a PR about read-only computed fields, but that is not totally related here.

So I just want to access a subset of the tables fields. Like I want to ignore the computed fields at all. Currently the code fails with a `KeyError` exception.

It should be possible to also only update the fields that were specified. The docs say:

> To update Event Datenbank records, issue a request to the Event Datenbank endpoint. Table names and table ids can be used interchangeably. Using table ids means table name changes do not require modifications to your API request. A PATCH request will only update the fields you specify, leaving the rest as they were. A PUT request will perform a destructive update and clear all unspecified cell values.

mesozoic:
I'm not sure there's a valid use case for the ORM to fail when it gets a field it doesn't recognize. If someone has a use case in mind, we can always make this a configuration flag on `Meta` later, but I think it should be permissive by default.

### DIFF
--- a/pyairtable/orm/model.py
+++ b/pyairtable/orm/model.py
@@ -216,7 +216,7 @@ class Model(metaclass=abc.ABCMeta):
             record = table.create(fields, typecast=self.typecast)
             did_create = True
         else:
-            record = table.update(self.id, fields, replace=True, typecast=self.typecast)
+            record = table.update(self.id, fields, typecast=self.typecast)
             did_create = False
 
         self.id = record["id"]
--- a/pyairtable/orm/model.py
+++ b/pyairtable/orm/model.py
@@ -260,15 +260,14 @@ class Model(metaclass=abc.ABCMeta):
         """Create instance from record dictionary"""
         name_attr_map = cls._field_name_attribute_map()
         name_field_map = cls._field_name_descriptor_map()
-        try:
-            # Convert Column Names into model field names
-            # Use field's to_internal to cast into model fields
-            kwargs = {
-                name_attr_map[k]: name_field_map[k].to_internal_value(v)
-                for k, v in record["fields"].items()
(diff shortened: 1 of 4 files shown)


pyairtable/orm/model.py
```diff
@@ -216,7 +216,7 @@ def save(self) -> bool:
             record = table.create(fields, typecast=self.typecast)
             did_create = True
         else:
-            record = table.update(self.id, fields, replace=True, typecast=self.typecast)
+            record = table.update(self.id, fields, typecast=self.typecast)
```
gtalarico, line 219: Can you explain rationale for removing replace?

gtalarico (approved):

pyairtable/orm/model.py
mesozoic: Sure! This one surprised me. Let's say we have a table with three fields: Name, Address, and Email. Let's also say our ORM only includes two of them:

```python
class Contact(Model):
    name = TextField("Name")
    address = TextField("Address")
```

Today, calling `Contact.from_id(value)` will fail if the record it retrieves has `Email` as a key. This branch fixes that. However, `Contact.from_id(value).save()` will _not_ pass a value for `Email` back to the API, which means passing `replace=True` would erase values that the ORM never declared.

I just pushed another commit with an integration test that shows exactly this behavior. This test fails with `replace=True`, but it passes if I remove that option.

gtalarico: Perfect, thank you for clarifying it. It's been a while since I have looked at this. Makes sense

PR #252, opened 2023-05-05
Fix return_fields_by_field_id on create/update, and add more tests

### LINKED ISSUE #194: Create and update commands do not take return_fields_by_field_id into account

mesozoic:
The reason this wasn't working before was that we were passing `returnFieldsByFieldId` as a query parameter, when we really needed to include it in the JSON payload in the POST body. This branch fixes that, adds some tests, and does a bit of tidying around the integration test suite.

On further digging, it doesn't look like [Create records](https://airtable.com/developers/web/api/create-records) or [Update records](https://airtable.com/developers/web/api/update-multiple-records) support _any_ of the options (besides this one) which we can pass to [List records](https://airtable.com/developers/web/api/list-records), so I removed the `**options` kwargs from those method signatures.

Fixes #194

### DIFF
--- a/pyairtable/api/abstract.py
+++ b/pyairtable/api/abstract.py
@@ -200,14 +200,22 @@ class ApiAbstract(metaclass=abc.ABCMeta):
         return all_records
 
     def _create(
-        self, base_id: str, table_name: str, fields: dict, typecast=False, **options
+        self,
+        base_id: str,
+        table_name: str,
+        fields: dict,
+        typecast=False,
+        return_fields_by_field_id=False,
     ):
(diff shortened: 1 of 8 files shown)


mesozoic:
`return_fields_by_field_id=` is a mouthful, and I'd probably have picked something more concise (like `field_ids=`). But I left it alone because I didn't want to break any code that already uses the existing parameter.

mesozoic:
Gonna merge now. If anyone spots a problem or wants to reopen a decision in here, just tag me and we'll talk it through.

PR #255, opened 2023-05-09
{Api,Base,Table}.batch_upsert and docs

### LINKED ISSUE #216: Upserts on table
The Airtable API is supporting upset functionality now. This feature is very useful when we want to update the records based on a specific key instead of system object ID. Can we implement this feature in here also?

mesozoic:
This adds support and documentation for the `performUpsert` mechanism described in [Update multiple records](https://airtable.com/developers/web/api/update-multiple-records).

One thing that might merit discussion: Before making any API calls, this implementation will validate that each record provided has the required "key fields", so that implementers don't have to worry about a single malformed record halting an incomplete operation (and we don't have to worry about how to represent that scenario). There are a lot of circumstances where I don't think we should be second-guessing what an implementer passes into the library; in this case we can _reliably_ predict a failure, so I think we should save them the trouble.

This would close #209 and resolve #216. @marks @dhanaschenniappan @marekvigas FYI

### DIFF
--- a/docs/source/api.rst
+++ b/docs/source/api.rst
@@ -150,6 +150,20 @@ Batch update records from a list of records.
   [{'id': 'recwPQIfs4wKPyc9D', 'fields': {"First Name": "Matt", ...}}, ...]
 
 
+:meth:`~pyairtable.api.Table.batch_upsert`
+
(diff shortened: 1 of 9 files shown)


marks:
Thanks so much @mesozoic and sorry I missed your comment on #209!

Issue #249, opened 2023-05-03
Roadmap for future releases


mesozoic:
The following is a rough proposal for which features we plan to add to pyAirtable, loosely grouped into releases. Priorities can shift over time, and anyone who wants to see a particular feature implemented faster is welcome to submit pull requests for review. I'm comfortable with the scope of the 1.5.0 and 2.0.0 releases; everything past there is a bit more speculative, but I have proofs of concept for at least some of it.

This document assumes we are using [semantic versioning](https://semver.org/), and that existing APIs can *only* be broken if there is a major version number change. As such, we may be more cautious about implementing new endpoints (or accepting PRs for new endpoints) if the design of that API merits discussion, because we'll commit to supporting new APIs for at least one major version.

Tagging other recent code contributors for input. @bapcon @goksan @larsakerson @nicohood

## 1.5 (released May '23)

This release will focus on bug fixes, quick wins, and reducing the open issue count.

- [x] Documentation typos and fixes (see [197](https://github.com/gtalarico/pyairtable/issues/197))
- [x] Implement continuous integration with type checking (see [195](https://github.com/gtalarico/pyairtable/issues/195), [217](https://github.com/gtalarico/pyairtable/issues/217), [239](https://github.com/gtalarico/pyairtable/issues/239), [240](https://github.com/gtalarico/pyairtable/issues/240), [241](https://github.com/gtalarico/pyairtable/issues/241), [242](https://github.com/gtalarico/pyairtable/issues/242))
- [x] #243
- [x] #182
- [x] #250 (see [190](https://github.com/gtalarico/pyairtable/issues/190))
- [x] #247 (see [203](https://github.com/gtalarico/pyairtable/issues/203), [210](https://github.com/gtalarico/pyairtable/issues/210), [222](https://github.com/gtalarico/pyairtable/issues/222))
- [x] #255 (see [209](https://github.com/gtalarico/pyairtable/issues/209), [216](https://github.com/gtalarico/pyairtable/issues/216))
- [x] #194

## 2.0 (July '23)

This release will be our chance to make any breaking API changes, such as changes to default behavior or return types. We will also implement stricter typing across the library, which might break CI for downstream code that does its own type checking, and support every field type in the ORM.

- [x] #261
- [x] #260
- [x] #262
- [x] Remove (or fully implement) `validate_type=` parameter to `Field` constructor
- [x] #263 (see [202](https://github.com/gtalarico/pyairtable/issues/202))
[...]

PR #256, opened 2023-05-10
Release 1.5.0

mesozoic:
I think we've got everything done that we said we would do in #249 for the 1.5.0 release. Here's an updated changelog and some "Added in 1.5.0" notes on the bits of the API that are new. 

I'll leave this open for the week to give us time to spot any last minute changes needed, and plan to push to PyPI on May 15.

### DIFF
--- a/docs/source/api.rst
+++ b/docs/source/api.rst
@@ -152,6 +152,8 @@ Batch update records from a list of records.
 
 :meth:`~pyairtable.api.Table.batch_upsert`
 
+.. versionadded:: 1.5.0
+
 Batch upsert (create or update) records from a list of records. For details on the behavior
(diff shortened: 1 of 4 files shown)


mesozoic:
Just to make sure I understand the existing flow, it looks like [the publish task](https://github.com/gtalarico/pyairtable/blob/main/.github/workflows/test_lint_deploy.yml#L42) in GitHub Actions will upload to PyPI if I push tag `1.5.0`, so I should run `make release-test` (but _not_ `make release`) on my local machine before pushing that tag. 

@gtalarico does that sound right?

gtalarico:
This looks great.

As far as release, it's been a while but that sounds about right.

gtalarico (approved):

Issue #258, opened 2023-05-15
For discussion: adding Pydantic for metadata models 


mesozoic:
This is a proposal for how to represent complex nested data structures from the Airtable API. This proposal would benefit from, but does not strictly require, removing ApiAbstract (see #257).

## tl;dr

I'd like to add [pydantic](https://docs.pydantic.dev/latest) as a dependency and use that to serialize and deserialize Airtable models from their metadata APIs.

## Rationale

Most Python developers these days use supportive development tools that can provide type hints, autocomplete, and more. Developers who need to interact with the nested data structures returned by the Airtable API would benefit from being able to navigate those within their code editors' tooling.

Some projects using this library might also want to enforce strict typing, and today there's no common way for them to ensure that the properties they reference on pyairtable's return types actually exist.

## Design

The module layout is _very_ open to discussion, but it could be something like this:

- `pyairtable/`
    - `metadata/` **(new)**
        - `field_schema.py`: Models for [each type of field config](https://airtable.com/developers/web/api/field-model).
        - `base_schema.py`: Models for base and table schemas, permissions, and invites.
        - `webhooks.py`: Models for interacting with webhooks.
    - `api/`
        - `api.py`: We might add some utility methods for accessing metadata APIs.
        - `enterprise.py` **(new)**: Define an `Enterprise` class and methods for managing users and retrieving audit logs.
        - `workspace.py` **(new)**: Define a `Workspace` class and methods for managing workspace permissions.
        - `base.py`: Add `Base` methods for retrieving and managing base schemas, permissions, and webhooks.
        - `table.py`: Add `Table` methods for retrieving and managing table/field schemas.

## Example

From a user's perspective, this will be relatively transparent. They will call methods that we expose on classes defined in `pyairtable.api`, and retrieve normal Python data structures that they can interact with:

```python
>>> table = Table("api_key", "appId", "tblId")
>>> schema = table.get_schema()
>>> type(schema)
<class 'pyairtable.metadata.base_schema.TableSchema'>
>>> schema.id
'tblId'
>>> schema.fields[0].id
'fld1VnoyuotSTyxW1'
>>> schema.fields[0].type
'singleLineText'
```

For now I'm not envisioning these data structures knowing how to call the API or save modifications to themselves. We can probably start with bespoke methods for each type of modification, for example:

```python
>>> table = Table("api_key", "appId", "tblId")
>>> table.get_schema().description
'Apartments to track.'
[...]

PR #263, opened 2023-05-20
Support `mypy --strict`

### LINKED ISSUE #202: mypy error: Module "pyairtable" does not explicitly export attribute "Api"; implicit reexport disabled
I get this when running `mypy --string` on my code:

```
test/init_test.py:16: error: Module "pyairtable" does not explicitly export attribute "Api"; implicit reexport disabled
test/init_test.py:16: error: Module "pyairtable" does not explicitly export attribute "Table"; implicit reexport disabled
```

The relevant line is just:

```
from pyairtable import Api, Table
```

This is how pymongo solved the same issue: https://jira.mongodb.org/browse/PYTHON-3311

mesozoic:
This branch adds type annotations to the entire library, along with a number of Airtable-specific types and `TypedDict`s. It is not compatible with Python 3.7 but it works on all later versions. As with some of the other big branches, it might be easier to review each commit individually.

* https://github.com/gtalarico/pyairtable/commit/821e435df80c4cf518116cb81032a7c73249ff68 introduces a number of types and `TypedDict`s in a new `pyairtable.api.types` module. 
    * There are some aliases here (like `RecordId` and `FieldName`) that simply point to `str`. This is not strictly necessary, but without them there are so many places where we have things like `List[str]` or `Dict[str, str]`, and this might make it a bit easier for developers to remember what goes where. 
    * I'm open to adding more but I didn't want to go overboard on the first pass. (I'm not _opposed_ to seeing `str` in type annotations as long as it's obvious what they represent.)

* https://github.com/gtalarico/pyairtable/commit/4aab9e077d3828bc5fd8c20de6ad49062f69a017 switches up how we call `mypy` to use strict mode and not silence any warnings.

* https://github.com/gtalarico/pyairtable/commit/5345b048cc1f56533b3bf73f039c737c037b3313 fills in most of the missing annotations on function/method signatures. It also removes the (deprecated) `params.to_params_dict` function and introduces `Model._get_meta()` as a helper for reading the `Meta` class.

* https://github.com/gtalarico/pyairtable/commit/8a28f3470988a0e758df0a1ca8bdc805bec6c054 does a few things:

* Turns `pyairtable.orm.fields.Field` into a generically-typed class, `Field[T_API, T_ORM]`, whose subclasses need to define the API and ORM data types (which might not always match). 
    * Provides the alias `BasicField[T]` for field classes which do not convert values between types. 
    * Provides the alias `AnyField` for when a type annotation doesn't care what kind of field it gets.
    * Renames `Field.value_if_missing` to `Field._missing_value()` so it can be typed more flexibly.

* https://github.com/gtalarico/pyairtable/commit/b618b55409c2306e66fba26687a2d3fd86fdb5c2 uses `typing.cast()` on values returned by the Airtable API to conform them to the types introduced above; this assumes the API responses will conform to the data structures we expect. (See below.)

* https://github.com/gtalarico/pyairtable/commit/3d9be4731a51605ea1dd581f91ee0bdc0f76a5ca fixes some issues with forward references that caused sphinx_autodoc_typehints to crash.

* https://github.com/gtalarico/pyairtable/commit/ffa5c431223f9a0bcde22edaf61147c80eabbea6 introduces some basic types to `pyairtable.metadata`. I expect to deprecate this module soon, so I didn't spend too much time on it.

* https://github.com/gtalarico/pyairtable/commit/46b0950750a3dd9021cbc87e30816de5a40ada05 removes support (introduced in an earlier branch) for setting `model.list_field = iterable_object`, where `iterable_object` is a non-list (such as a set, tuple, or generator). Defining this correctly as a type annotation became problematic, and it seems reasonable to expect developers to convert those types to lists _before_ setting field values on a model.

* https://github.com/gtalarico/pyairtable/commit/8364a4f916f33d077383851828152199c980ae58 adds a special test file, `tests/test_typing.py`, which makes a number of type assertions that mypy will validate on each run.

* https://github.com/gtalarico/pyairtable/commit/94a93cda63237d3220aae0a6a25a1443d17f87eb changes the approach used in b618b554 (above) to actively validate Airtable's API responses against our TypedDict definitions, instead of simply assuming they match what we expect. This introduces [pydantic](https://docs.pydantic.dev/latest/) as a dependency, which we've already discussed using for metadata models in #258.

* 34eb19f updates the Sphinx configuration to include type annotations and default values in function signatures and parameter descriptions. It also adds a "Types" section.

* 9b5ecd0 fixes a bug that popped up when I pushed the previous commit, but which I hadn't run across before. I'm stumped why GitHub Actions didn't catch it when I pushed 94a93cda but :shrug: it was a quick fix.

This branch resolves #202.

### DIFF
--- a/docs/source/api.rst
+++ b/docs/source/api.rst
@@ -411,3 +411,10 @@ can use if you want to compose formulas:
 .. autofunction:: pyairtable.formulas.IF
 .. autofunction:: pyairtable.formulas.STR_VALUE
 .. autofunction:: pyairtable.formulas.LOWER
+
+
+Types
+*******
+
+.. automodule:: pyairtable.api.types
+    :members:
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -3,7 +3,9 @@
 #
 import os
 import sys
+import typing
 
+import pyairtable.api.types
 from pyairtable import __version__ as version
 
 extensions = [
@@ -22,19 +24,6 @@ extensions = [
 # Add any paths that contain templates here, relative to this directory.
 templates_path = ["_templates"]
 
-# Auto Create Api Reference
-autoapi_dirs = ["../pyairtable"]
-autoapi_add_toctree_entry = True
-autoapi_options = [
-    "members",
-    "undoc-members",
-    "private-members",
-    "show-inheritance",
-    "show-module-summary",
-    "special-members",
-    "imported-members",
-]
-
 # Open Graph extension config. https://pypi.org/project/sphinxext-opengraph/
 ogp_site_url = "https://pyairtable.readthedocs.io/"
 ogp_image = "https://pyairtable.readthedocs.io/en/master/_images/logo.png"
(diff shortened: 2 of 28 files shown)


mesozoic:
Going to merge this now. Please feel free to open further issues for discussion if anything looks awry.

PR #267, opened 2023-05-30
Drop `ApiAbstract`

### LINKED ISSUE #257: For discussion: consider removing `ApiAbstract`
This is a proposal for discussion. It's not a "must have". I think it's worth considering prior to a major version release, since 2.0.0 is our best opportunity to make potentially breaking API changes.

## tl;dr

I think we should consider removing `ApiAbstract`. I think the library API would be simpler for implementers to understand if it favored composition over inheritance. I think this will make it easier to add more endpoints in the future.

## Why change it?

I'm suggesting this change to the structure of the API because:

1. It involves a lot of duplication. Adding one method (like #250) involves adding three other methods which call the abstract class via `super()`. We need to keep the signatures and API documentation for each of those methods up-to-date.

2. This will not scale gracefully if we want to add dozens of endpoints. Keeping four separate method signatures up-to-date for every action we might take (update a webhook, post a comment, etc) will become burdensome over time.

3. The current approach exposes methods which don't make sense. Take the following code snippet as an example:

    ```python
    >>> table = Table("key", "app", "tbl")
    >>> table.get_table_url()
    Traceback (most recent call last):
      File "<stdin>", line 1, in <module>
    TypeError: ApiAbstract.get_table_url() missing 2 required positional arguments: 'base_id' and 'table_name'
    ```

    ...the correct thing to do is call `table.table_url` rather than `table.get_table_url()`, but the presence of both (a byproduct of inheritance) is confusing clutter, especially if you're using an IDE that does autocompletion.

4.  There are a number of endpoints (particularly around metadata) that only ever make sense in the context of a base, or a table, or a record; keeping those methods scoped to just the one appropriate class will simplify maintenance _and_ make it easier for implementers to understand their use.

## How would this work?

This is just a proposal, but one way I might arrange this is:

- [ ] **ApiAbstract** goes away
- [ ] **Api** contains basic utilities for connecting to the Airtable API, and nothing more.
  - [x] `__init__(access_token)`
  - [x] `build_url(...) -> str`
  - [ ] `build_request(...) -> requests.Request`
  - [ ] `prepare_request(...) -> requests.PreparedRequest`
  - [x] `request(...) -> dict` executes a request and returns the deserialized JSON payload.
  - [x] `base(base_id) -> Base` (renamed from `get_base`)
  - [ ] `table(base_id, table_name) -> Table` (renamed from `get_table`)
  - [ ] [`create_base`](https://airtable.com/developers/web/api/create-base) (to be implemented)
- [ ] **Base** is where we'd add future methods for interacting with [schemas](https://airtable.com/developers/web/api/get-base-schema), [webhooks](https://airtable.com/developers/web/api/webhooks-overview), etc.
  - [x] `__init__(api, base_id)` could take an instance of `Api` or `str`
  - [x] `api`, `url` would be available as properties
  - [x] `table(table_name) -> Table` (renamed from `get_table`)
  - [ ] [`create_table`](https://airtable.com/developers/web/api/create-table) (to be implemented)
- [ ] **Table** would contain a lot of the methods that exist in ApiAbstract today.
  - [x] `__init__(access_token: str, base_id: str, table_name: str)` (old style constructor)
  - [x] `__init__(base: Base, table_name: str)` (new style constructor)
  - [x] `api`, `base`, `url` would be available as properties
  - [x] `record_url(record_id) -> str` (renamed from `get_record_url`)
  - [x] `get(record_id: str, **options)`
  - [x] `iterate(**options)`
  - [x] `first(**options)`
  - [x] `all(**options)`
  - [x] `create(fields: dict)`
  - [x] `batch_create(records)`
  - [x] `update(record_id: str, fields: dict)`
  - [x] `batch_update(records: List[dict])`
  - [x] `batch_upsert(records: List[dict], key_fields: List[str])`
  - [x] `delete(record_id: str)`
  - [x] `batch_delete(record_ids: List[str])`
  - [ ] [`update_schema`](https://airtable.com/developers/web/api/update-table) (to be implemented)
  - [ ] [`create_field`](https://airtable.com/developers/web/api/create-field) (to be implemented)
  - [ ] [`update_field`](https://airtable.com/developers/web/api/update-field) (to be implemented)

So the most typical use cases would still be supported:

```python
>>> table = Table("token", "appWhatever", "tblWhatever")
>>> records = table.all(formula=formula)
>>> # do stuff
```

...but could also be achieved through composition, which might be useful in some circumstances (depending on how the implementer is using this library):

```python
>>> api = Api("token")
>>> table = Table(api, "appWhatever", "tblWhatever")
>>> records = table.all(formula=formula)
```

...or perhaps like this:

```python
>>> api = Api("token")
>>> table = api.base("appWhatever").table("tblWhatever")
>>> records = table.all(formula=formula)
```

There would be less confusion over inherited methods that don't quite apply:

```python
>>> table.get_table_url()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: 'Table' object has no attribute 'get_table_url'
>>> table.url
'https://api.airtable.com/v0/appWhatever/tblWhatever'
```

...and there would be less boilerplate required every time we need to add a method, since we'd only ever add it in one place.

## Why _shouldn't_ we do this?

I can think of two big reasons:

1. Breaking API changes are never _fun_ and maybe we just don't want the hassle. I think this can be mitigated by (A) getting the constructors to take old-style arguments (`str`s) as well as new-style arguments (instances), and (B) using wrappers that emit deprecation warnings to preserve some of the old method names for a period of time.

2. It's not strictly necessary. We could start adding lots of things to the existing `Base` or `Table` classes and _not_ add them to the `ApiAbstract` base class. We'd be breaking an existing pattern, though, and it would be somewhat arbitrary which methods existed in all places and which ones only existed where they "belong". There would also still be some methods in `ApiAbstract` which don't belong in subclasses (like `get_table_url`).

I'm open to hearing about more reasons not to consider this :grin:

mesozoic:
This closes #257 by removing `ApiAbstract`. Instead, we have `Api`, `Base`, and `Table` classes that each have separate functions and responsibilities. They do still refer to each other, and their constructors have at least some backwards compatibility. See below (or read the changes to `migrations.rst`) for more details.

This will be a breaking API change, so definitely want to get eyes on this before I merge it. Tagging recent contributors for ideas or feedback. @BAPCon @goksan @larsakerson @NicoHood @marks 

## A note on circular dependencies

`Api` needs to know how to create `Base` and `Table` instances, but the `Base` and `Table` instances need to have references back to the `Api` and `Base` (respectively). To avoid a circular reference, these modules import each other using fully qualified names, and only refer to the classes within methods that need them. Type annotations are enclosed in quotes.

This results in internal code that might be a bit aesthetically displeasing, like:

```python
class Base:
    def table(self, table_name):
        return pyairtable.api.table.Table(table_name)

class Table:
    api: "pyairtable.api.api.Api"
    base: "pyairtable.api.base.Base"
```

This was, however, the simplest form I could find that allows for both type annotations and bidirectional runtime dependencies. `Api` needs to be able to create a `Base`, but for backwards compatibility, we also need `Base` to be able to create an `Api`.

## A note on backwards compatibility

The most common use case for this library (based on a [cursory search](https://github.com/search?q=pyairtable+-repo%3Agtalarico%2Fpyairtable+%2F+%3D+%28Api%7CBase%7CTable%29%5C%28%2F+language%3APython&type=code&l=Python)) seems to be:

```python
table = Table(api_key, base_id, table_name)
```

This branch preserves backwards compatibility with that constructor syntax, but will emit deprecation warnings implying that this style could be removed in the future:

```python-repl
>>> from pyairtable import Base, Table
>>> base = Base("api_key", "base_id")
<stdin>:1: DeprecationWarning: Passing API keys to pyairtable.Base is deprecated; use Api.base() instead. See https://pyairtable.rtfd.org/en/latest/migrations.html for details.
>>> table = Table("api_key", "base_id", "table_name")
<stdin>:1: DeprecationWarning: Passing API keys or base IDs to pyairtable.Table is deprecated; use Api.table() or Base.table() instead. See https://pyairtable.rtfd.org/en/latest/migrations.html for details.
```

The rationale for deprecating this code path is that, in the future, we might want to perform additional steps when referencing a table (such as validating its existence). This will be easier to do in `Base.table()` _before_ constructing the `Table` instance.

## Summary of each commit

- 1a90f8b removes `ApiAbstract` and moves its methods into either `Api`, `Base`, or (mostly) `Table`. I tried not to change behavior or introduce any new methods, but did add two: 
  - `Api.chunked()` breaks an iterable into chunks based on Airtable's maximum record count per request
  - `Api.wait()` sleeps for 1/N seconds, where N is Airtable's QPS limit. (I hope to remove the need for this soon.)
- 46232cd started a section of `migrations.rst` that discusses how to upgrade.
- ac5e654 is a substantial reorganization of docs/ to reflect the new Api structure.
- c144f68 removes anything from `substitutions.rst` that was unused, or only used in one place.
- f210d4a removes the middle column from the `Parameters` section (which isn't strictly necessary) so that the rest of the table is easier to read.
- 6285dff removes the "Added in 1.0.0" label; now that we are on 2.0.0, it doesn't feel necessary to reference everything that was added when this library was renamed.
- 5313443 folds the "utils" docs into the API Reference section.
- 0f8548e adds sections to the `changelog.rst` and `migrations.rst` for the 2.0 release
- e523a5c changes the constructors for the refactored `Base` and `Table` classes so that they can receive either `str` or instances in their constructors. (See backwards compatibility, above.)
- 3e146ee improves backwards-compatibility of the `Table()` constructor so that it can accept Api kwargs when receiving all `str` arguments.
- c1c536b is another pass over the docs to improve legibility and fix typos.
- 98a336a standardizes references within our docs to the top-level name (e.g. `pyairtable.Table`) instead of the canonical name (e.g. `pyairtable.api.table.Table`)


mesozoic:
I'm working on some other branches to finish up the rest of the 2.0 release scope, but they depend on this one so I'm not posting them yet. I'll plan to merge this branch next week if there are no objections or questions raised here.

PR #268, opened 2023-06-06
Allow ORM Meta options to be lookups at runtime (not import time)

### LINKED ISSUE #192: ORM specify configuration and API key outside the code
Hi,
I have a simple ORM Model and of course I dont want to place any API keys inside the code, and I want to make those settings configurable anyways (as I have special testing and live databases on airtable).

```py
class AirtableEvent(airtable.Model):
    title = airtable.fields.TextField("Titel")
    description = airtable.fields.TextField("Beschreibung")
    start = airtable.fields.DatetimeField("Start")
    end = airtable.fields.DatetimeField("Ende")
    url = airtable.fields.TextField("Url")
    address = airtable.fields.TextField("Adresse")
    location = airtable.fields.TextField("Ort")
    status = airtable.fields.TextField("Status")
    recurring = airtable.fields.CheckboxField("Wiederkehrend")
    source = airtable.fields.TextField("Quelle")
    google_calendar_meta = airtable.fields.TextField("Google Calendar Meta")
    google_uid_meta = airtable.fields.TextField("Google Uid Meta")
    google_sequence_meta = airtable.fields.IntegerField("Google Sequence Meta")
    trash = airtable.fields.CheckboxField("Papierkorb")

    class Meta:
        # This must be loaded from the configuration
        base_id = None
        table_name = None
        api_key = None

    @classmethod
    def load_json_config(cls, api_key, config):
        cls.Meta.base_id=config["base_id"]
        cls.Meta.table_name=config["table_id"]
        cls.Meta.api_key=api_key
```

Now if you use the class as it is right now, it will fail with `ValueError: Meta.base_id must be defined in model`

I can surpress this by setting it to a non-empty string and set those metadata afterwards. But how can we do this in a more proper way? I am searching for a good solution, do you have any idea?

mesozoic:
This allows the `Meta` class inside each ORM model to provide a callable that will be executed whenever the ORM requests one of its meta properties. The use case we have is to be able to provide API keys via global configuration singletons or environment variables which might not be loaded at import time but will be present at runtime.

Resolves #192

### DIFF
--- a/pyairtable/orm/model.py
+++ b/pyairtable/orm/model.py
@@ -20,9 +20,9 @@ from pyairtable.orm.fields import AnyField, Field
 
 class Model:
     """
-    This class allows you create an orm-style class for your Airtable tables.
+    This class allows you create an ORM-style class for your Airtable tables.
 
-    This is a meta class and can only be used to define sub-classes.
+    This is a metaclass and can only be used to define sub-classes.
 
     The ``Meta`` is reuired and must specify all three attributes: ``base_id``,
     ``table_id``, and ``api_key``.
@@ -38,6 +38,22 @@ class Model:
     ...         api_key = "keyapikey"
     ...         timeout: Optional[Tuple[int, int]] = (5, 5)
     ...         typecast: bool = True
+
+    You can implement meta attributes as callables if certain values
+    need to be dynamically provided or are unavailable at import time:
+
+    >>> from pyairtable.orm import Model, fields
+    >>> class Contact(Model):
+    ...     first_name = fields.TextField("First Name")
+    ...     age = fields.IntegerField("Age")
+    ...
+    ...     class Meta:
+    ...         base_id = "appaPqizdsNHDvlEm"
+    ...         table_name = "Contact"
+    ...
+    ...         @staticmethod
+    ...         def api_key():
+    ...             return os.environ["AIRTABLE_API_KEY"]
     """
 
     id: str = ""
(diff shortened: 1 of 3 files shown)


pyairtable/orm/model.py
```diff
@@ -171,6 +187,8 @@ def _get_meta(cls, name: str, default: Any = None, required: bool = False) -> An
         if required and not hasattr(cls.Meta, name):
             raise ValueError(f"{cls.__name__}.Meta.{name} must be defined")
         value = getattr(cls.Meta, name, default)
+        if callable(value):
```
taylorwilsdon, line 190: Beyond a nit and this particular usage is already very concise so feel free to completely disregard, PR looks great but was poking around in Python 3.10 docs and this seems like a case where match (ref [PEP 634/635/636](https://peps.python.org/pep-0636/) could be used in a world where 3.10 is the minimum supported version ie:

```
match value:
    case callable_func := callable(value):
        value = callable_func()
```

taylorwilsdon (approved):

pyairtable/orm/model.py
mesozoic: I will admit to not having my head wrapped around the `match` syntax (or [soft keywords](https://docs.python.org/3/reference/lexical_analysis.html#soft-keywords) in general) but I think it's fair to say we shouldn't use it if it's not present (or available via `__future__`) in all supported Python versions.

PR #272, opened 2023-06-15
Use a default retry strategy and remove `time.sleep`

mesozoic:
This implements a default retry strategy which will retry 429 (too many requests) up to five times before giving up. This seems like a sensible default, and it is similar to the behavior of the official Airtable.js library. The brief testing I've done suggests that removing `time.sleep` and replacing with retries does speed up large batch operations. 

Given this code...
```python
import os
import time
from threading import Thread
from pyairtable import Api


def main():
    for case in [(1, 500), (5, 100), (25, 20)]:
        timer(*case)


def timer(threads, records):
    start_time = time.time()
    started = []
    for thread_number in range(threads):
        t = Thread(target=bulk_create, args=[thread_number, records])
        t.start()
        started.append(t)

    all(t.join() for t in started)
    duration = time.time() - start_time
    print(threads, "thread(s),", records, "records:", f"{duration:.2f}s")


def bulk_create(prefix, count):
    api = Api(os.environ["AIRTABLE_API_KEY"])
    table = api.table("appaPqizdsNHDvlEm", "TEST_TABLE")
    records = [{"text": f"bulk {prefix} #{n}"} for n in range(1, count)]
    created = table.batch_create(records)
    table.batch_delete([record["id"] for record in created])


if __name__ == "__main__":
    main()
```

...I got this output:
```
% git checkout main
% python3 tmp/bulk.py
1 thread(s), 500 records: 53.67s
5 thread(s), 100 records: 12.96s
Exception in thread Thread-14:
...
requests.exceptions.HTTPError: 429 Client Error: Too Many Requests

% git checkout default_retry
% python3 tmp/bulk.py
1 thread(s), 500 records: 32.18s
5 thread(s), 100 records: 10.19s
25 thread(s), 20 records: 10.02s
```

I'd considered retrying 500s by default, but I think that is probably inadvisable given a 500 might indicate a request was _partially_ fulfilled. It should be left to the implementer to decide whether to retry an operation in that situation or not.

This does not add any new tests. I considered adding something like the above as an integration test, but it would add several seconds to each test run and I'm not clear on what sort of regressions we'd actually be guarding against.

This branch closes #218

### DIFF
--- a/docs/source/api.rst
+++ b/docs/source/api.rst
@@ -34,14 +34,23 @@ Retrying
 `urllib3.util.Retry <https://urllib3.readthedocs.io/en/stable/reference/urllib3.util.html#urllib3.util.Retry>`_.
 
 As of 2.0.0, the default behavior is to retry requests up to five times if the Airtable API responds with
-a 429 status code (indicating you've exceeded their per-base QPS limit) or a potentially transient server error.
+a 429 status code, indicating you've exceeded their per-base QPS limit.
 
-To adjust the default behavior, you can use the :func:`~pyairtable.retry_strategy` function:
+To adjust the default behavior, you can use the :func:`~pyairtable.retry_strategy` function.
+For example, to increase the total number of retries:
 
 .. code-block:: python
 
   from pyairtable import Api, retry_strategy
   api = Api('auth_token', retry_strategy=retry_strategy(total=10))
+
+Or to retry certain types of server errors in addition to rate limiting errors:
+
+.. code-block:: python
+
+  from pyairtable import Api, retry_strategy
+  retry = retry_strategy(status_forcelist=(429, 500, 502, 503, 504))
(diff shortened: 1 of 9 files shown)


mesozoic:
It seems this code only works with urllib3 >= 1.26, when they changed the signature of the `Retry` class, whereas the minimum version we test against is 1.25.11. Given that 1.26.0 was released in 2020, I'm going to suggest we bump up the urllib3 dependency for this package instead of trying to remain backwards-compatible.

gtalarico (approved):

mesozoic:
I added an integration test that does (more or less) what the batch above does; it spawns 25 threads and makes multiple calls to `batch_create` from each of them. Without the default retry strategy, the test fails. With it, the test passes in ~14 seconds. Better to be confident we won't regress this in the future.

Issue #275, opened 2023-06-16
version 1.5.0 - pip installs different version of API from what's on Github


kaiwenfa:
Seems like the version of pyairtable when installed from pip is different from what's on the Github and shown in the documentation.

For example, when tried to retrieve a table by calling `api.table(baseId, tableId)`, I got `AttributeError: 'Api' object has no attribute 'table'`. after looking into the source code in the installed pyairtable package, I saw in the `Api` class there's no `table()` method, but `get_table()` instead.

mesozoic:
Thanks for reporting this! It seems like the issue has to do with the way we're generating our documentation. https://pyairtable.readthedocs.io/en/latest is rendering the documentation as they exist in `main`, which is a work in progress prior to release. It really should be pointing to the documentation for the `1.5.0` tag, which is the most recent released version. (We don't update the version numbers in GitHub until we're ready to release.)

mesozoic:
Sorry about the confusion, folks! You were seeing future documentation for an unreleased version. I've changed our configuration on readthedocs.org so that the default version is `stable` (which currently points to `1.5.0`) instead of `latest` (which points to `main`). 

I've also hidden the `latest` docs entirely to avoid any further confusion, at the cost of potentially breaking some URLs. I'll activate it again once we've released 2.0, because we don't expect more breaking API changes after that.

I noticed that my browser cache was causing me to get reloaded to the old page for a while, so if you find that https://pyairtable.rtfd.org is still redirecting you to `latest`, use this URL: https://pyairtable.readthedocs.io/en/stable/

@gtalarico FYI

mesozoic:
On further digging, it looks like I _can_ force the `latest` docs to point to `1.5.0` instead of `main`, so I'm going to do that and reactivate the `latest` docs. That way we can avoid breaking any URLs, and we just need to remember to set this setting back when we release 2.0.

ericrobskyhuntley:
Thanks, all!

ericrobskyhuntley:
Thanks so much, @mesozoic !

kaiwenfa:
Thanks for looking into this!

PR #276, opened 2023-06-19
Update getting-started.rst

suryasanchez:
Update the documentation to fix:
AttributeError: 'Api' object has no attribute 'table'


mesozoic:
Hi there! This seems to be related to #275. The docs in `main` are correct for the code in `main`, but the docs on readthedocs.org are "from the future" and need to be adjusted. I'm on it, thanks for the report!

PR #277, opened 2023-06-25
Move ORM out of "experimental" status

mesozoic:
This branch attempts to bring the ORM module out of experimental status and into a stable API, including full support for type annotations. There are lots of edge cases with an ORM, though, so I'm open to ideas on other tests we should write first. This is hopefully the last big branch before we can get ready for the 2.0 release. 

Here is a list of the major commits in this branch; this list skips a few minor cleanup/tidying commits that are not worth highlighting.

* b9d6145 is a substantial reorganization of the documentation, including:
    * A more concise "Getting Started" section
    * A longer section on interacting with data that brings together tables, formulas, and retries
    * A dedicated page for the ORM, with some auto-generated content using [cog](https://nedbatchelder.com/code/cog)
    * A more-or-less complete API reference on a single page
* 69f0434 introduces type validation on `LinkField` to ensure we aren't accidentally sending record IDs to the API from the wrong table.
* 628a58d adds a very basic integration test for a new table that has every single type of field on it, including errors and NaN results. I could've done a lot more here, but it wasn't clear it was necessary and I want to minimize dependence on integration tests vs. unit/mock tests.
* cb5e2f7 and a81a9e7 add support for dynamic resolution of model classes for linked fields.
    * `model="some.module.Model"` will use `importlib` to import `Model` from `some.module`
    * `model="Model"` will import `Model` from the same module where the field's model is defined
    * `model=LinkSelf` will ensure the `LinkField` points back to the same model where it is defined
* 6c1135d removes the `_linked_cache` property on the Model class. 
    * I spent a bit of time writing tests for this behavior, for things like "a record which links to itself should return the same object from the link field". It was fun but challenging. I wound up with something very brittle, non-thread-safe, and with a ton of edge cases and guesswork. 
    * Ultimately I don't think this is something the ORM should try to do. Let implementers deal with caching if it matters to them.
    * This commit also changes where in the data flow we fetch linked records; instead of fetching them at instance creation (when calling `to_internal_value()` on each field), we fetch them the first time `field.__get__` is called.
* e8cda82 ensures that any list field (LinkField, MultipleSelectField, etc.) always returns an empty list (never `None`) and that its type annotations reflect that. All other fields (TextField, CollaboratorField, etc.) are still annotated as `Optional`, to remind implementers that the value could be `None`.

Open questions that I haven't bottomed out on myself:

- [x] Should we change the default for `LinkField` to be `lazy=False`? Now that we don't load data until the field's `__get__` is called (see 6c1135d) it seems less useful to default to lazy loading fields.
- [ ] Are there other use cases for the ORM that we don't have captured in our regression tests? Esoteric combinations of Airtable field configuration that can produce data structures we don't recognize?
- [x] Should we do more around error handling? There are certain fields (like rollups or formulas) that can return errors or NaNs, but our type hints don't make that apparent and there's no specific handling for it. We provide `is_airtable_error()` but maybe that's not enough.


gtalarico (approved):
Nice to see the orm mature into stable release.
Amazing work as usual 👏

PR #278, opened 2023-06-30
Release 2.0.0rc1

mesozoic:
A few last minute cleanups on documentation and a full list of PRs in the changelog. I think this is ready to rumble. 

I am inclined to leave this open until after the US holiday, in case anyone out there has random ideas/feedback to share, or in case someone in the world has an unpinned dependency and we're about to ruin their build. 😬


mesozoic:
As part of testing this release against production code, I found that enforcing types on `RecordDict` values became pretty chaotic, especially in edge cases (like lookups with errors). I think it's probably a mistake for us to define types on record responses, since Airtable could add new response types at any point. I changed this in 8009cedbb. We do still type check what an implementer writes to the API, but we _don't_ type check what comes back from the API.

PR #281, opened 2023-07-19
Return full payload from batch_upsert

### LINKED ISSUE #280: batch_upsert does not return createdRecords or updatedRecords
I believe I have the code pretty right, but when I am trying to batch_upsert records I am not getting the createdRecords in the response. Anything look wrong?

```
    def upsert_record(self, records):
        try:
            return self.table.batch_upsert(records=records,key_fields=['field_name'],return_fields_by_field_id=True)
            print('upsert success')
        except Exception as E :
            print('upsert failed')
            print(E)
```

mesozoic:
* Changes the return signature of `batch_upsert` from a list of records to a dict
* Iterates over multiple pages of inputs and combines them all into a single response
* Fixes #280 – FYI @awb715

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -7,6 +7,8 @@ Changelog
 
 See :ref:`Migrating from 1.x to 2.0` for detailed migration notes.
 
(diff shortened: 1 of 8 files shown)


awb715:
Thank you for this fix!

PR #288, opened 2023-08-01
Workaround for pydantic v2

xl0:
It's a bit heavier than expected to make mypy happy on both pydantic versions.

### DIFF
--- a/pyairtable/api/types.py
+++ b/pyairtable/api/types.py
@@ -3,18 +3,11 @@ pyAirtable provides a number of type aliases and TypedDicts which are used as in
 and return values to various pyAirtable methods.
 """
 from functools import lru_cache
-from typing import TYPE_CHECKING, Any, Dict, List, Optional, Type, TypeVar, Union, cast
-
-if TYPE_CHECKING:  # mypy really does not like this conditional import.
-    import pydantic
-else:
-    # Pydantic v2 broke a bunch of stuff. Luckily they provide a built-in v1.
-    try:
-        import pydantic.v1 as pydantic
-    except ImportError:
-        import pydantic
+from typing import Any, Dict, List, Optional, Type, TypeVar, Union, cast
 
 from typing_extensions import Required, TypeAlias, TypedDict
+
+from pyairtable._compat import pydantic
 
 T = TypeVar("T")
 
--- a/pyairtable/api/types.py
+++ b/pyairtable/api/types.py
@@ -3,9 +3,17 @@ pyAirtable provides a number of type aliases and TypedDicts which are used as in
 and return values to various pyAirtable methods.
 """
 from functools import lru_cache
-from typing import Any, Dict, List, Optional, Type, TypeVar, Union, cast
+from typing import TYPE_CHECKING, Any, Dict, List, Optional, Type, TypeVar, Union, cast
 
-import pydantic
+if TYPE_CHECKING:  # mypy really does not like this conditional import.
+    import pydantic
+else:
+    # Pydantic v2 broke a bunch of stuff. Luckily they provide a built-in v1.
+    try:
+        import pydantic.v1 as pydantic
+    except ImportError:
+        import pydantic
+
 from typing_extensions import Required, TypeAlias, TypedDict
 
 T = TypeVar("T")
@@ -287,7 +295,8 @@ def _create_model_from_typeddict(cls: Type[T]) -> Type[pydantic.BaseModel]:
     Creates a pydantic model from a TypedDict to use as a validator.
     Memoizes the result so we don't have to call this more than once per class.
     """
-    return pydantic.create_model_from_typeddict(cls)
+    # Mypy can't tell that we are using pydantic v1.
+    return pydantic.create_model_from_typeddict(cls)  # type: ignore[no-any-return, operator, unused-ignore]
 
 
 def assert_typed_dict(cls: Type[T], obj: Any) -> T:
--- a/pyairtable/models/_base.py
+++ b/pyairtable/models/_base.py
@@ -1,8 +1,17 @@
 from functools import partial
-from typing import Any, ClassVar, Iterable, Optional
+from typing import TYPE_CHECKING, Any, ClassVar, Iterable, Optional
 
 import inflection
-import pydantic
+
+if TYPE_CHECKING:  # mypy really does not like this conditional import.
+    import pydantic
+else:
+    # Pydantic v2 broke a bunch of stuff. Luckily they provide a built-in v1.
+    try:
+        import pydantic.v1 as pydantic
+    except ImportError:
+        import pydantic
+
 from typing_extensions import Self as SelfType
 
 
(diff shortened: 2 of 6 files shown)


pyairtable/models/_base.py
```diff
@@ -86,7 +95,8 @@ def deleted(self) -> bool:
 
     def __setattr__(self, name: str, value: Any) -> None:
         # Prevents implementers from changing values on readonly or non-writable fields.
-        if name in self.__class__.__fields__:
+        # Mypy can't tell that we are using pydantic v1.
+        if name in self.__class__.__fields__:  # type: ignore[operator, unused-ignore]
```
mesozoic, line 99: I was able to drop the `unused-ignore` portion of this without reintroducing any failures.

tests/test_api_types.py
```diff
@@ -1,5 +1,15 @@
+from typing import TYPE_CHECKING
+
 import pytest
-from pydantic import ValidationError
+
+if TYPE_CHECKING:  # mypy really does not like this conditional import.
+    import pydantic
```
mesozoic, line 6: Should this be `import pydantic.v1 as pydantic`, since mypy is going to run in a venv with `pydantic>=2`?

I'm not sure it winds up making a difference because we now have `# type: ignore` everywhere, but it seems "more correct", and perhaps it will make a difference at some point in the future.

pyairtable/api/types.py
```diff
@@ -3,9 +3,17 @@
...
+from typing import TYPE_CHECKING, Any, Dict, List, Optional, Type, TypeVar, Union, cast
+
+if TYPE_CHECKING:  # mypy really does not like this conditional import.
+    import pydantic
+else:
+    # Pydantic v2 broke a bunch of stuff. Luckily they provide a built-in v1.
+    try:
+        import pydantic.v1 as pydantic
+    except ImportError:
+        import pydantic
```
mesozoic, line 15: Should we move this somewhere (maybe `pyairtable.utils` or a new module `pyairtable.compat`) so it only happens once, and then import from there in the rest of our code?

mesozoic (changes requested):
Fantastic! 🙌  Thank you for figuring this out! 

I like this approach because it is agnostic which version our library users require for other parts of their project. Also, since it does not require changing any of our APIs or behaviors, we can make it a patch release instead of a minor version bump.

I suggested a couple changes below. I also think we could add a few lines to `tox.ini` to ensure that our test suite runs successfully against both major versions of pydantic:

```diff
diff --git a/tox.ini b/tox.ini
index 584ae4d..142db2e 100644
--- a/tox.ini
+++ b/tox.ini
@@ -3,6 +3,7 @@ envlist =
     pre-commit
     mypy
     py3{8,9,10,11}-requests{min,max}
+    py3{8,9,10,11}-pydantic1
     coverage
 
 [gh-actions]
@@ -29,6 +30,7 @@ deps =
     -r requirements-test.txt
     requestsmin: requests==2.22.0  # Keep in sync with setup.cfg
     requestsmax: requests>=2.22.0  # Keep in sync with setup.cfg
+    pydantic1: pydantic<2  # Lots of projects still use 1.x
 
 [testenv:coverage]
 passenv = COVERAGE_FORMAT
```

I'll be very happy to merge this branch once we've got test coverage and have at least tried those other ideas.

pyairtable/models/_base.py
xl0: Are you sure you tried with v2? I think v2's BaseModel does not have __fields__, so mypy complains here.

tests/test_api_types.py
xl0: This would cause a mypy error for v1 users, would it not?

pyairtable/api/types.py
xl0: I think this is a natural reaction to seeing duplicated code, but I feel in this case, the abstraction will just make it harder to follow what's going on.

xl0:
Agree re tox, I'll add it later today.

pyairtable/api/types.py
xl0: Actually, since you are adding more pedantic, it definitely makes sense to move this into one place. I will do it.

pyairtable/models/_base.py
mesozoic: 🤷  I'm running `tox -e mypy` and it's using the latest pydantic version, but I'm not sure it's worth our time to keep tweaking this particular nuance. Having `# type: ignore` feels kinda icky regardless, so who cares which options it needs? 😁

tests/test_api_types.py
mesozoic: Hmm, I was (selfishly) thinking that since we run `tox -e mypy` in a separate environment, it will always use the latest pydantic. But you're right, we also have to think about library users who are running mypy on their own projects (and might have pinned dependencies to pydantic v1 _or_ v2).

I'm uncertain of the best way to handle this, but we probably want to test this branch as a dependency within some live code (that uses mypy) before we merge it. I can try to get to that in a week or so, but if you can sooner, all the better.

pyairtable/api/types.py
mesozoic: Thanks! I think most library users are not going to peek under the hood at how we're using pydantic; they're just going to want to import `Base` and `Table` and get on with life. So if it makes it easier for us (and I think it will, given how many places I'm starting to use pydantic) it's worth the time.

pyairtable/models/_base.py
xl0: Yeah, it needs ignore-unused for v1 because the warning only occurs with pydantic v2.

xl0:
@mesozoic I updated the PR to
- Add -pydantic1 to tox tests. Also added mypy-pydantic1.
- Moves pydantic stuff into pyairtable/pydantic.py. Can't be in utils due to circular dependency.

xl0:
@mesozoic , any idea what's up with the docs failing?

mesozoic:
> @mesozoic , any idea what's up with the docs failing?

I think you might've forgotten to `git add` the new file. Can we call it `_compat.py`, though, to (1) make it clear that it's internal and not part of the public API, and (2) to avoid any confusion arising from the module and its contents having the same name?

xl0:
@mesozoic , heh, my bad. Ok, renamed it to _compat.py.

tests/test_api_types.py
mesozoic: I used this branch in a project of mine with both `pydantic~=1` and `pydantic~=2`, and the test suite (including mypy) passed in both cases. 🎉

mesozoic (approved):

mesozoic:
@xl0 One more request, could you rebase the branch with [signed commits](https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification)? Easiest way is with SSH, in my experience, but the repo requires it before we merge. I could rebase and merge it on your behalf, but then you won't get any credit. 😁

xl0:
Ok, I'll check it later today.

xl0:
@mesozoic , done

mesozoic:
Thanks very much for solving this!

PR #291, opened 2023-08-06
Webhooks support

mesozoic:
This branch introduces models representing the [webhooks API](https://airtable.com/developers/web/api/webhooks-overview) along with methods for validating notifications and for retrieving new webhook payloads. It adds the following methods and classes to the public API:

* `Base.add_webhook(url, spec)`
* `Base.webhooks()`
* `Base.webhook(id)`
* `models.WebhookNotification`
* `models.WebhookNotification.from_request(body, header, secret)`
* `models.Webhook`
* `models.Webhook.{enable,disable}_notifications()`
* `models.Webhook.extend_expiration()`
* `models.Webhook.payloads(cursor=1)`
* `models.WebhookPayload`

The major design decision here is to _not_ return plain old `dict` data structures from the API, and instead to parse them into nested pydantic data models with type annotations, using `extra = "ignore"` so that it doesn't crash if Airtable adds more fields in the future. 

The advantage of this approach is a cleaner API and more assistance for developers who use type-aware IDEs; the disadvantage is that any new fields which Airtable adds to their API in the future will require a patch release to fully support. I've worked around this by adding a private attribute called `_raw` to every model which will be the API payload which generated the model. In a pinch, a developer could access that directly if it contains anything our models don't recognize.

Tagging a few folks for input:

* @gtalarico for code review and guidance
* @marks in case Airtable has imminent changes planned to these APIs
* @xl0 because of recent interest in use cases for pydantic
* @BAPCon because this resolves #236


marks:
Hi @mesozoic - thanks for doing this! I took a quick skim and it looks good to me. 

No imminent changes and I think it's smart to ensure additional fields do not break types/etc as [the deprecation guidelines note](https://support.airtable.com/docs/airtable-api-deprecation-guidelines#what-do-we-mean-by-breaking-change-and-deprecation):
> Note that adding new data in a forwards-compatible way is not considered a breaking change (e.g. adding new keys and values to an object returned from an API endpoint).

PR #301, opened 2023-08-19
Fix docs bug that treated all `Dict[str,Any]` as `Fields`

mesozoic:
Unfortunately in https://github.com/gtalarico/pyairtable/commit/8009cedbb7d7524e81ab6c46f0667f66f73333c8 it seems I inadvertently broke a part of the custom type hints formatter in `conf.py`, which relied on type aliases generally being unique. This caused it to display any `Dict[str, Any]` as a link to the `Fields` alias, which is subtly confusing.

For now I don't have a great workaround. The plugin we're using to render type annotations in method parameters doesn't support the autodoc_type_aliases option (see https://github.com/tox-dev/sphinx-autodoc-typehints/issues/284). Instead I'm just going to limit our use of it to a couple specific type aliases which we know  are unique.

I also found a couple spots where we were using `Fields` instead of `WritableFields`. This means we're _technically_ changing the API, but I doubt it's in a way that will impact anyone's project. I think we can release this as 2.1.1 at some point (instead of waiting for 2.2).

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -70,8 +70,11 @@ def typehints_formatter(annotation, config):
         if isinstance(value, type) and issubclass(value, dict):  # TypedDict
             return f":data:`~pyairtable.api.types.{name}`"
         # At this point there's no way to determine whether an annotation
-        # which was evaluated to Dict[str, Any] is Fields or something unrelated.
-        # TODO: get all TypeAlias annotations properly reflected in docs.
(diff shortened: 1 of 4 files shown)


mesozoic:
I also found that [autodoc-pydantic v2](autodoc-pydantic.readthedocs.io) doesn't work with pydantic v1 models, so I had to pin that dependency to get back field summaries, etc. for Pydantic models.

PR #306, opened 2023-09-13
Fix documentation

NicoHood:
I guess that should be the correct wording, right?


mesozoic:
Good catch! Thanks for the patch. 🙌

PR #310, opened 2023-09-29
Support for "AI Text" field type

mesozoic:
I haven't actually needed this or used it, I'm just going off what the documentation says. FYI @marks in case there's some nuance to these fields that I haven't captured here.


SeanKeenan-at:
Sean here (engineer that implemented this public API interface on Airtable's side 👋 ).

No nuance that you're missing! We'd ideally support more functionality in the future (eg: writing these cell values or triggering AI generation), but do not have support for that via the public API yet.

Thanks for adding this!

pyairtable/api/types.py
```diff
@@ -25,6 +25,25 @@
...
+    {
+        'state': 'generated',
+        'isStale': False,
+        'value': '...'
+    }
+    """
+
+    state: Required[str]
+    isStale: Required[bool]
+    value: Required[Optional[str]]
```
SeanKeenan-at, line 43: The only type I'm not sure about is this - I assume this means string | null, but I'm not familiar with Python types 😄

mesozoic: Recognizing the absurdity of having to write "required and optional", you've got it exactly right. If we enabled [postponed annotations](https://docs.python.org/3/whatsnew/3.7.html#pep-563-postponed-evaluation-of-annotations), we could change this to `Required[str | None]`.

Issue #307, opened 2023-09-19
Update does not accept return_fields_by_field_id parameter


NicoHood:
Same issue as https://github.com/gtalarico/pyairtable/issues/194

This PR should have fixed it:
https://github.com/gtalarico/pyairtable/pull/252

But the parameter is still missing:
https://pyairtable.readthedocs.io/en/stable/_modules/pyairtable/api/table.html#Table.update

mesozoic:
That's true! Unfortunately this is a limitation of the Airtable API; passing `returnFieldsByFieldId` to the update endpoint returns a 422 (unlike the batch update endpoint, which does the right thing). 

In the branch you mentioned, there's a [commented-out test](https://github.com/gtalarico/pyairtable/blob/2e4f3b896739c5d7e80f339c19a8b58d9d14f5d0/tests/integration/test_integration_api.py#L68) showing how we _could_ test it, if it were supported.

NicoHood:
WTF? Alright, wouldnt it make sense to use the batch api then? I mean we could keep the same syntax and whenever airtable fixes this issue we could switch the call internally? That would make more sense, as we do not need to work around in the user code. Or is there any advantage in the "normal", single update call?

marks:
Hi @NicoHood & @mesozoic - [the Airtable PATCH/PUT endpoints for a single record](https://airtable.com/developers/web/api/update-record) [now](https://airtable.com/developers/web/api/changelog#anchor-2023-10-03) support `returnFieldsByFieldId`

NicoHood:
Nice! So then we could just enable it, right?

mesozoic:
I'll reopen the issue. Pull requests are welcome!

NicoHood:
I am sorry I aint got time to work on this. I know the fix is easy, but I have other priorities. Would be nice if someone else could add this parameter :-)

marks:
@NicoHood - I believe PR #320 takes care of this for you if you want to go over and subscribe for updates over there

mesozoic:
Thanks for resolving this issue, @marks!

PR #320, opened 2023-11-25
Add return_fields_by_field_id to Table.update()

marks:
As requested and discussed in #307


marks:
BTW - maintainers - is there documentation for running integration tests locally? I was able to figure out what I needed to create in Airtable in tests/integration/conftest.py and related tests, but just curious if I had missed something

mesozoic:
Thanks for adding this, @marks! Right now the repo requires all commits to be signed; so that you get credit for the PR, could you [enable signed commits](https://docs.github.com/articles/about-gpg/) and rebase your branch?

We do not have any documentation for the integration tests; that's a todo I never got around to. They should work out-of-the-box with creator privileges on the test base, and I just sent you an invite for that. However, the non-integration test suite should still provide as close to 100% test coverage as possible.

marks:
Ah, thank you @mesozoic !

All commits for this branch/PR should be signed/verified now.

PR #323, opened 2023-12-09
Add formulas for missing logical operators

MichalKacprzak99:
Implements missing formulas from [section "Logical operators and functions" from documentation of "Formula field reference"](https://support.airtable.com/docs/formula-field-reference#logical-operators-and-functions). Currently only `EQUAL` is implemented.

### DIFF
--- a/pyairtable/formulas.py
+++ b/pyairtable/formulas.py
@@ -218,3 +218,53 @@ def LOWER(value: str) -> str:
     "LOWER(TestValue)"
     """
     return "LOWER({})".format(value)
+
+
+def NOT_EQUAL(left: Any, right: Any) -> str:
+    """
+    Create an inequality assertion
+
+    >>> NOT_EQUAL(2,2)
+    '2!=2'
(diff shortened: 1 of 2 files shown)


mesozoic:
I know this follows the existing pattern, but I'll admit that I'm not sure the current layout of the `formulas` module adds a lot of value. It doesn't strike me as being any easier to type `formulas.GREATER_THAN(a, b)` than the equivalent `f"{a} > {b}"`. 

I'm working on a branch that will (among other things) apply `to_airtable_value` to all function parameters, so if they're decimals or dates or some other type that needs Airtable-specific formatting, it's not the responsibility of the caller. But that branch is going to be a breaking change for a 3.0 release, because there's no real way to do it in a totally backwards-compatible manner.

I'm going to merge this branch because tests pass and it follows existing convention, but I will probably still propose breaking changes to this module for the next major release.

mesozoic:
Thanks for contributing!

Issue #343, opened 2024-03-12
Error on `base.create_table`


BAPCon:
Not sure what exactly is going on here, but when I call the following code it throws a `key error`. I am looking into it further, but incase I forget I figured I'd leave an issue open. I think it has something to do with a schema not being ready (???) but validation is being set to `true` when returning the created table.

```
base = api.base(base_name)
tables = base.tables()
base.create_table(
    name="table_name",
    description="Hello world!",
    fields=[
    {
        "description": "Name of the apartment",
        "name": "Name",
        "type": "singleLineText"
    },
    {
        "name": "Address",
        "type": "singleLineText"
    },
    {
        "name": "Visited",
        "options": {
            "color": "greenBright",
            "icon": "check"
        },
        "type": "checkbox"
    }
]
)
```

```
Traceback (most recent call last):
  File "c:\Users\user\pyairtable\app.py", line 7, in <module>
    base.create_table(
  File "c:\Users\user\pyairtable\pyairtable\api\base.py", line 163, in create_table
    return self.table(response["id"], validate=True)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "c:\Users\user\pyairtable\pyairtable\api\base.py", line 120, in table
    schema = self.schema(force=force).table(id_or_name)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "c:\Users\user\pyairtable\pyairtable\models\schema.py", line 268, in table
    return _find(self.tables, id_or_name)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "c:\Users\user\pyairtable\pyairtable\models\schema.py", line 48, in _find
    return items_by_name[id_or_name]
           ~~~~~~~~~~~~~^^^^^^^^^^^^
KeyError: 'tblrxZtA9ld4bePi9'
```

BAPCon:
Yep. In `base.create_table()` at the line `return self.table(response["id"], validate=True)`, setting validate to `False` causes no error to be thrown.

PR #345, opened 2024-03-12
Update `base.create_table`

BAPCon:
Turn on `force` during `base.create_table` when trying to return the newly created table. The newly created table is not included in the cached tables, setting `force` to `True` makes sure to use the updated table list.

I believe this is necessary, as adding a `force` parameter to the arguments list of `base.create_table` does not make sense. I can't see any utility in not refreshing the tables list, unless we set returning the new table to optional.

See issue #343

### DIFF
--- a/pyairtable/api/base.py
+++ b/pyairtable/api/base.py
@@ -160,7 +160,7 @@ class Base:
         if description:
             payload["description"] = description
         response = self.api.post(url, json=payload)
-        return self.table(response["id"], validate=True)
+        return self.table(response["id"], validate=False)
 
     @property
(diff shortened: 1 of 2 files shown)


mesozoic:
Good catch - thanks for the bug report! Per our contribution guidelines could you add something to our regression tests which (1) fails on the `main` branch and (2) passes with this change?

BAPCon:
This should do it, though I am not super confident in my test writing ability lol. Let me know if more changes are needed

mesozoic:
Thanks for contributing! I'll get a bugfix release out today.

PR #347, opened 2024-03-14
Return `""` and `False` from TextField, CheckboxField

mesozoic:
It doesn't really make sense for a TextField or CheckboxField to return `None`, since Airtable will return `null` for a text field with an empty value or a checkbox field that is unchecked. This branch allows those fields to return empty strings or `False` as their "missing value", which will reduce the amount of defensive `if model.field is None: ...` style of type checking required by implementers.

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -5,6 +5,10 @@ Changelog
 3.0 (TBD)
 ------------------------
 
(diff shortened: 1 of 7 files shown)


mesozoic:
Adding to 3.0 milestone since this would technically be a breaking API change

Issue #314, opened 2023-11-02
get_table_schema fails when the api.table is initialized with the table's ID


calebclimatecabinet:
For example
```python
get_table_schema(api.table(base_id, "tblXXXXXXXXXXXXXX"))
```
returns `None` but the function works if you use the table's name:
```python
get_table_schema(api.table(base_id, "My Table"))
```

mesozoic:
Good catch. I can see that function is comparing to `name` but not to `id`. 

#311 introduces a new set of APIs for retrieving metadata, so I'm generally not inclined to do much maintenance on the existing code (which will be deprecated). But this should be easy enough to fix. Pull requests welcome 😁  otherwise I'll get to this the next chance I have.

calebclimatecabinet:
Ah no worries – if it'll be depreciated I'm fine letting it slide.

If anyone else comes across this I just wrote my own `get_table_schema` which only works in the opposite case:

```python
def get_table_schema(table):
    base_schema = get_base_schema(table)
    return next(t for t in base_schema["tables"] if t["id"] == table.name)
```

PR #349, opened 2024-03-15
Bug fix for deprecated `metadata.get_table_schema`

BAPCon:
Fixes the bug described in #314
Adds test case in `tests\test_api_table`


mesozoic:
Do we need to fix this bug? I'd planned on removing this module in the 3.0 release, and `Table.schema()` offers equivalent functionality without this limitation.

BAPCon:
Not sure, I figured since #314 was still open and had been updated/labeled as `bug` that we were looking to fix it. I figure if the deprecated feature is still active that a bugfix make sense, but it's up to you!

mesozoic:
Can't hurt to close the ticket, then. I'll merge it and do one more bugfix release before 3.0. Thanks for contributing!

PR #350, opened 2024-03-15
Add base.duplicate_table method 

BAPCon:
Not sure if this is a feature we want, but seeing as it can be done in the web UI I figured it may be worth adding.

Adds test function emulating duplicating a table with and without copying records.
The schema is always duplicated.

### DIFF
--- a/docs/source/metadata.rst
+++ b/docs/source/metadata.rst
@@ -68,6 +68,9 @@ The following methods allow creating bases, tables, or fields:
 .. automethod:: pyairtable.Base.create_table
     :noindex:
 
+.. automethod:: pyairtable.Base.duplicate_table
+    :noindex:
+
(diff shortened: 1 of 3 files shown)


mesozoic:
I'm not sure this is something that belongs in the client library since it's not a supported endpoint. Airtable does provide a web UI for copying a base, but they don't expose that functionality through the API. The "copy base" feature on the web has certain features, like copying comments, which we wouldn't be able to replicate ourselves, and some library users might be misled into thinking that this method offers equivalent functionality.

We've had some of our own use cases for things like this, and we've found there are a lot of detailed decisions to make that are specific to our own use case. I suspect other library users who might want to duplicate bases programmatically will also need to work through a lot of details on their own.

BAPCon:
Sounds good!

PR #355, opened 2024-03-20
Add attribute `use_field_ids` to `orm.Model.Meta`

### LINKED ISSUE #187: ORM specify fields as id, not as name
It should be possible to make use of the `return_fields_by_field_id` option with ORM as well

### LINKED ISSUE #351: ORM table `kwargs` wrapping
In regards to issues like #187 and #186 , are we looking to wrap records related fields in the ORM module? Not sure if those are issues we are actively looking to solve.

You can already achieve, for example, `return_fields_by_field_id` when invoking functions in the `Model` like `cls.all()` and `cls.first()`. However, that requires the user to repeat the `kwargs` on each call. Would an explicit approach that allows these options to be assigned in the `Meta` subclass fit into the overall roadmap?

```
@classmethod
    def _kwargs_union(cls, request_args: dict):
        meta_keys = [
            'view', 
            'fields', 
            'sort', 
            'formula', 
            'cell_format', 
            'user_locale', 
            'time_zone', 
            'return_fields_by_field_id'
        ]
        args_union = {attr: cls._get_meta(attr) for attr in meta_keys if cls._get_meta(attr)}
        args_union.update(request_args)
        return args_union
```

```
@classmethod
    def all(cls, **kwargs: Any) -> List[SelfType]:
        """
        Retrieve all records for this model. For all supported
        keyword arguments, see :meth:`Table.all <pyairtable.Table.all>`.
        """
        table = cls.get_table()
        kwargs = cls._kwargs_union(kwargs)
        return [cls.from_record(record) for record in table.all(**kwargs)]
```

This then allows:

```
class Contact(Model):
    name = fields.TextField("fldxqagV60dzo00fV")
    class Meta:
        base_id = "base"
        table_name = "table"
        api_key = "your_key"
        timeout = (5, 5)
        typecast = True
        return_fields_by_field_id=True
```

BAPCon:
Resolves #351
Resolves #187 

Adds wrapper `@modify_kwargs` that handles kwarg sanitization for:
- `Model.all()`
- `Model.first()`

`cell_format` and `return_fields_by_field_id` are removed.

If `Model.Meta.use_field_ids=True`, adds `return_fields_by_field_id=True` to above function calls.

I figure adding a dedicated wrapper is easier to read and is easier to manage if more sanitization is needed in the future.

### DIFF
--- a/pyairtable/orm/model.py
+++ b/pyairtable/orm/model.py
@@ -1,4 +1,4 @@
-from functools import lru_cache, wraps
+from functools import lru_cache
 from typing import Any, Dict, Iterable, List, Optional
 
 from typing_extensions import Self as SelfType
@@ -172,19 +172,8 @@ class Model:
                 )
             )
 
-    def modify_kwargs(func):
-        # Modifies kwargs passed to Model.all and Model.first
-        # If Meta.use_field_ids, kwargs['return_fields_by_field_id']=True.
-
-        @wraps(func)
-        def wrapper(cls, *args, **kwargs):
-            kwargs = cls._kwargs_all_first(kwargs)
-            return func(cls, *args, **kwargs)
-
-        return wrapper
-
     @classmethod
-    def _kwargs_all_first(cls, request_args: dict):
+    def _get_meta_request_kwargs(cls, request_args: dict):
         # Called by modify_kwargs to modify kwargs passed to `all()` & `first()`.
         disallowed_args = (
             "cell_format",
@@ -214,7 +203,6 @@ class Model:
     @classmethod
     def _typecast(cls) -> bool:
         _ = bool(cls._get_meta("typecast", default=True))
-        print(f"Typecast: {_}")
         return _
 
     def exists(self) -> bool:
@@ -265,22 +253,22 @@ class Model:
         return bool(result["deleted"])
 
     @classmethod
-    @modify_kwargs
     def all(cls, **kwargs: Any) -> List[SelfType]:
         """
         Retrieve all records for this model. For all supported
         keyword arguments, see :meth:`Table.all <pyairtable.Table.all>`.
         """
+        kwargs.update(cls._get_meta_request_kwargs(kwargs))
         table = cls.get_table()
         return [cls.from_record(record) for record in table.all(**kwargs)]
 
     @classmethod
-    @modify_kwargs
     def first(cls, **kwargs: Any) -> Optional[SelfType]:
         """
         Retrieve the first record for this model. For all supported
         keyword arguments, see :meth:`Table.first <pyairtable.Table.first>`.
         """
+        kwargs.update(cls._get_meta_request_kwargs(kwargs))
         table = cls.get_table()
         if record := table.first(**kwargs):
             return cls.from_record(record)
--- a/pyairtable/orm/model.py
+++ b/pyairtable/orm/model.py
@@ -176,7 +176,6 @@ class Model:
     def _get_meta_request_kwargs(cls):
         # Called by modify_kwargs to modify kwargs passed to `all()` & `first()`.
         return {
-            "cell_format": "json",
             "user_locale": None,
             "time_zone": None,
             "return_fields_by_field_id": (
--- a/pyairtable/orm/model.py
+++ b/pyairtable/orm/model.py
@@ -1,4 +1,4 @@
-from functools import lru_cache
+from functools import lru_cache, wraps
 from typing import Any, Dict, Iterable, List, Optional
 
 from typing_extensions import Self as SelfType
@@ -173,26 +173,25 @@ class Model:
             )
 
     def modify_kwargs(func):
-        """Decorator to modify the kwargs by calling _kwargs_union method."""
-        from functools import wraps
+        # Modifies kwargs passed to Model.all and Model.first
+        # If Meta.use_field_ids, kwargs['return_fields_by_field_id']=True.
 
         @wraps(func)
         def wrapper(cls, *args, **kwargs):
-            # Modify kwargs by calling the _kwargs_union method on the class.
-            kwargs = cls._kwargs_union(kwargs)
+            kwargs = cls._kwargs_all_first(kwargs)
             return func(cls, *args, **kwargs)
 
         return wrapper
 
     @classmethod
-    def _kwargs_union(cls, request_args: dict):
+    def _kwargs_all_first(cls, request_args: dict):
+        # Called by modify_kwargs to modify kwargs passed to `all()` & `first()`.
         disallowed_args = (
             "cell_format",
             "return_fields_by_field_id",
         )
         args_union = {k: v for k, v in request_args.items() if k not in disallowed_args}
-        use_field_ids = cls._get_meta("use_field_ids")
-        if use_field_ids:
+        if cls._get_meta("use_field_ids"):
             args_union["return_fields_by_field_id"] = True
         return args_union
 
--- a/pyairtable/orm/model.py
+++ b/pyairtable/orm/model.py
@@ -177,11 +177,22 @@ class Model:
         @wraps(func)
         def wrapper(cls, *args, **kwargs):
             # Modify kwargs by calling the _kwargs_union method on the class.
-            if hasattr(cls, "_kwargs_union"):
-                kwargs = cls._kwargs_union(kwargs)
+            kwargs = cls._kwargs_union(kwargs)
             return func(cls, *args, **kwargs)
 
         return wrapper
+
+    @classmethod
+    def _kwargs_union(cls, request_args: dict):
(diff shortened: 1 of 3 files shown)


pyairtable/orm/model.py
```diff
@@ -170,6 +172,29 @@ def _validate_class(cls) -> None:
...
+    def modify_kwargs(func):
+        # Modifies kwargs passed to Model.all and Model.first
+        # If Meta.use_field_ids, kwargs['return_fields_by_field_id']=True.
+
+        @wraps(func)
+        def wrapper(cls, *args, **kwargs):
+            kwargs = cls._kwargs_all_first(kwargs)
+            return func(cls, *args, **kwargs)
+
+        return wrapper
```
mesozoic, line 184: This shouldn't be inside the Model class – though I'm not sure it's really necessary. The `all` and `first` methods can do something like this and it's only one line of code:

```python
def all(self, **kwargs) -> ...:
    kwargs.update(self._get_meta_request_kwargs())
    ...
```

pyairtable/orm/model.py
```diff
@@ -170,6 +172,29 @@ def _validate_class(cls) -> None:
...
+
+    @classmethod
+    def _kwargs_all_first(cls, request_args: dict):
```
mesozoic, line 187: I would call this something like `_get_meta_request_kwargs` to keep the namespace tidy.

pyairtable/orm/model.py
```diff
@@ -188,7 +213,9 @@ def get_table(cls) -> Table:
 
     @classmethod
     def _typecast(cls) -> bool:
-        return bool(cls._get_meta("typecast", default=True))
+        _ = bool(cls._get_meta("typecast", default=True))
+        print(f"Typecast: {_}")
+        return _
```
mesozoic, line 218: Let's drop the debug statement :)

pyairtable/orm/model.py
```diff
@@ -170,6 +172,29 @@ def _validate_class(cls) -> None:
...
+    def _kwargs_all_first(cls, request_args: dict):
+        # Called by modify_kwargs to modify kwargs passed to `all()` & `first()`.
+        disallowed_args = (
+            "cell_format",
+            "return_fields_by_field_id",
+        )
+        args_union = {k: v for k, v in request_args.items() if k not in disallowed_args}
+        if cls._get_meta("use_field_ids"):
+            args_union["return_fields_by_field_id"] = True
+        return args_union
```
mesozoic, line 196: If we follow the pattern above, I think we could probably simplify this down to something like:

```
kwargs = {
    "cell_format": "json",
    "user_locale": None,
    "time_zone": None,
    "return_fields_by_field_id": cls._get_meta("use_field_ids"),
}
```

I also wonder if this would be a good place to put `kwargs["fields"] = [...]` so that we don't retrieve more data from the API than our ORM actually requires.

mesozoic, line 196: If we follow the pattern above, I think we could probably simplify this down to something like:

```
return {
    "cell_format": "json",
    "user_locale": None,
    "time_zone": None,
    "return_fields_by_field_id": cls._get_meta("use_field_ids"),
}
```

I also wonder if this would be a good place to put `kwargs.setdefault("fields", ...)` so that we don't retrieve more data from the API than our ORM actually requires. Right now the ORM will retrieve values for all fields on a table, regardless of whether they're mapped to a field descriptor.

tests/test_orm_model.py
```diff
@@ -193,6 +193,27 @@ def test_passthrough(methodname):
 
     mock_endpoint.assert_called_once_with(a=1, b=2, c=3)
 
+    
+def test_disallowed_args():
+    """
+    Test the argument sanitizer, `use_field_ids` attribute, and disallowed kwargs
+    """
```
mesozoic, line 200: This looks like it exercises everything; however, since it's heavily mocked, we're never actually testing that a return payload using field IDs (instead of names) will be mapped correctly to the ORM's fields. Maybe we can find a way to do that?

mesozoic (changes requested):
Thanks for contributing a fix! I like this idea and I think it'll make the ORM more useful. I have a few suggestions below but otherwise I'm looking forward to adding this.

pyairtable/orm/model.py
BAPCon: I thought we were going to outright remove the ability to pass "return_fields_by_field_id" without sanitizing the kwarg keys it allows those keys through.

I think we might have to then change some tests, I am getting:
`FAILED tests/test_orm.py::test_undeclared_field[first] - TypeError: int() argument must be a string, a bytes-like object or a real number, not 'NoneType'
FAILED tests/test_orm.py::test_undeclared_field[all] - TypeError: int() argument must be a string, a bytes-like object or a real number, not 'NoneType'`

BAPCon: Unless I change test_orm_model.py:
```
@pytest.mark.parametrize("methodname", ("all", "first"))
def test_passthrough(methodname):
    """
    Test that .all() and .first() pass through whatever they get.
    """
    with mock.patch(f"pyairtable.Table.{methodname}") as mock_endpoint:
        method = getattr(FakeModel, methodname)
        method(a=1, b=2, c=3)

    mock_endpoint.assert_called_once_with(a=1, b=2, c=3, cell_format="json", return_fields_by_field_id=False)
```

mesozoic: That change looks reasonable to me. Where an Airtable API parameter has one obvious "right answer", it's probably better for orm.Model to be explicit about what it expects.

tests/test_orm_model.py
BAPCon: Let me know if the testing works, I had to add  a separate `fake_meta_from_ids` and a separate model `FakeModelByIds`. If there is more professional way to do this let me know?

pyairtable/orm/model.py
```diff
@@ -46,6 +47,7 @@ class Meta:
                 api_key = "keyapikey"
                 timeout = (5, 5)
                 typecast = True
+                use_field_ids = True
```
mesozoic, line 50: Let's leave this out of the example, since the sample code (above) uses field names and not IDs. We want the example to remain internally consistent.

pyairtable/orm/model.py
```diff
@@ -170,6 +172,20 @@ def _validate_class(cls) -> None:
                 )
             )
 
+    @classmethod
+    def _get_meta_request_kwargs(cls):
+        # Called by modify_kwargs to modify kwargs passed to `all()` & `first()`.
```
mesozoic, line 177: Comment is out-of-date; let's drop this line

pyairtable/orm/model.py
```diff
@@ -170,6 +172,20 @@ def _validate_class(cls) -> None:
...
+        return {
+            "user_locale": None,
+            "cell_format": "json",
+            "time_zone": None,
+            "return_fields_by_field_id": (
+                cls._get_meta("use_field_ids")
+                if cls._get_meta("use_field_ids")
+                else False
```
mesozoic, line 185: I think this can be simplified to `cls._get_meta("use_field_ids", default=False)`

pyairtable/orm/model.py
```diff
@@ -188,7 +204,8 @@ def get_table(cls) -> Table:
 
     @classmethod
     def _typecast(cls) -> bool:
-        return bool(cls._get_meta("typecast", default=True))
+        _ = bool(cls._get_meta("typecast", default=True))
+        return _
```
mesozoic, line 208: Let's revert this change

pyairtable/testing.py
```diff
@@ -42,6 +42,24 @@ def fake_meta(
...
+    """
+    Generate a ``Meta`` class for inclusion in a ``Model`` subclass.
+    """
+    attrs = {
+        "base_id": base_id,
+        "table_name": table_name,
+        "api_key": api_key,
+        "use_field_ids": use_field_ids,
+    }
+    return type("Meta", (), attrs)
```
mesozoic, line 60: Do we need a new function for this, or can we add a **kwargs parameter to the existing `fake_meta`?

tests/test_orm_model.py
```diff
@@ -207,7 +254,9 @@ def test_dynamic_model_meta():
 
     class Fake(Model):
         class Meta:
-            api_key = lambda: data["api_key"]  # noqa
+            def api_key():
+                return data["api_key"]  # noqa
+
```
mesozoic, line 259: I don't think we needed to change this, did we?

mesozoic (changes requested):
This is looking good! The test seems to satisfy all the conditions we need to test, and it looks like we've maintained 100% coverage. A couple more pieces of feedback here, but nothing major.

mesozoic (approved):
Looks great! Thank you for contributing a new feature!

PR #362, opened 2024-04-01
Rename `return_fields_by_field_id=` to `use_field_ids=`

mesozoic:
This is a breaking change for 3.0 and it's totally cosmetic and silly, but I find the name "return fields by field id" to be way too much of a mouthful. We called this `use_field_ids` in #355 and I'd like to rename the kwarg to match. When users upgrade the package it should be easy enough to search/replace everywhere. I am comfortable being told this is a bad idea anyway.


gtalarico:
@mesozoic +1
That always bother me as well

PR #369, opened 2024-04-22
Reintroduce support for memoizing linked models

mesozoic:
In 6c1135dff251a39c078c9c1faa160a17b874a77a we removed the behavior of memoizing references to existing linked model instances because it was hard to test and led to unpredictable behavior. However, it is actually _very useful_ to prefetch all linked models rather than grabbing them one by one via API calls, because it can be much faster.

This branch reintroduces memoization as an optional feature, with more granular control over how and when it happens.

### DIFF
--- a/docs/source/_substitutions.rst
+++ b/docs/source/_substitutions.rst
@@ -69,6 +69,20 @@
     If ``True``, will fetch information from the metadata API and validate the ID/name exists,
     raising ``KeyError`` if it does not.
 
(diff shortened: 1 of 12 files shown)


mesozoic:
Good catch :)  I've updated tox.ini so that it runs mypy locally on every version of Python. Previously that only happened when running GitHub Actions.

jorge-yelp (approved):

Issue #378, opened 2024-08-01
Model.save() specific fields


gildastone:
Is there a way to save specific fields of a Model and not the entire record while using the `Model.save` method?

Indeed, I have a record that has a pretty heavy payload (number of fields). I'm fetching the record, doing some operations/updates on it on **specific fields** and then save again.

I would like to only save/update those specific fields instead of posting the whole record (see [pyairtable.orm.model.Model.save()](https://github.com/gtalarico/pyairtable/blob/7029671178f2964848104fde3e13e09d37a20edc/pyairtable/orm/model.py#L237C9-L237C62)):

I couldn't find a way with the current ORM. Is there a solution?

If not, we could update the `pyairtable.orm.model.Model.save` method to (adding the `fields_to_save: List = []` argument):

```python
    def save(self, fields_to_save: List = []) -> bool:
        if self._deleted:
            raise RuntimeError(f"{self.id} was deleted")
        table = self.meta.table
        fields = self.to_record(only_writable=True)["fields"]

        if fields_to_save:
            fields = {key: value for key, value in fields.items() if key in fields_to_save}

        if not self.id:
            record = table.create(fields, typecast=self.meta.typecast)
            did_create = True
        else:
            record = table.update(self.id, fields, typecast=self.meta.typecast)
            did_create = False

        self.id = record["id"]
        self.created_time = datetime_from_iso_str(record["createdTime"])
        return did_create
```

mesozoic:
Totally agree – this would be a very useful behavior (and should probably be the default behavior, rather than something implementers would have to explicitly request with a flag or a separate method call).

One way I can think of approaching this is to keep track of which values have been modified in memory since the object was created. Then it would be possible to only send those fields to the `table.update` method. There are a few good pieces of prior art for [SQLAlchemy](https://docs.sqlalchemy.org/en/20/orm/extensions/mutable.html), [Django](https://django-dirtyfields.readthedocs.io/en/stable/), and [Rails](https://api.rubyonrails.org/classes/ActiveModel/Dirty.html).

Pull requests welcome! 😁

BAPCon:
@mesozoic Working on this now, pull request probably today.

BAPCon:
@mesozoic I addressed the original issue and the move to being able to specify which fields to update, I believe my pull request accomplishes what the user stated.

[...]

PR #379, opened 2024-08-15
Enhance `Model.save` Method to Allow Field-specific Updates

BAPCon:
Introduces enhancements requested in #378 to the `Model.save` method, allowing specific field updates to a record rather than the entire record.

```python

contact = Contact(
    first_name="Benjamin",
    last_name="Perkins"
)

# Specify fields to save
contact.save(fields=['first_name'])

Contact.from_id(contact.id)._fields
>>> {'First Name': 'Benjamin'}
```

### ORM attribute name vs Airtable field name
I think the fields should be specified via their ORM attribute name: `first_name`, `last_name`. This would maintain consistency and be more Pythonic.

### DIFF
--- a/pyairtable/orm/model.py
+++ b/pyairtable/orm/model.py
@@ -230,7 +230,9 @@ class Model:
 
         Args:
             fields: A list of field names to save. If provided, only these
-                fields will be updated. Otherwise, all fields will be updated.
+            fields will be updated. Otherwise, all fields will be updated.
+            The field names should be in the form the user declared them
+            in the ORM class.
 
         Returns:
             ``True`` if a record was created, ``False`` if it was updated.
--- a/pyairtable/orm/model.py
+++ b/pyairtable/orm/model.py
@@ -246,7 +246,11 @@ class Model:
         if fields:
             # Convert ORM attribute names to Airtable field names
             update_fields = {
-                attribute_map[field].field_name: None if self._fields.get(attribute_map[field].field_name) is None else attribute_map[field].to_record_value(self._fields.get(attribute_map[field].field_name))
+                attribute_map[field].field_name: None
+                    if self._fields.get(attribute_map[field].field_name) is None 
+                    else attribute_map[field].to_record_value(
+                        self._fields.get(attribute_map[field].field_name)
+                    )
                 for field in fields
                 if field in attribute_map and not attribute_map[field].readonly
             }
--- a/pyairtable/orm/model.py
+++ b/pyairtable/orm/model.py
@@ -345,9 +345,9 @@ class Model:
         ct = datetime_to_iso_str(self.created_time) if self.created_time else ""
         return {"id": self.id, "createdTime": ct, "fields": fields}
     
-    def to_update_record(self, fields: Optional[List[str]] = None) -> UpdateRecordDict:
+    def to_update_record(self, fields: Optional[List[str]] = None) -> RecordDict:
         """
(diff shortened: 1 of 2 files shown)


pyairtable/orm/model.py
```diff
@@ -221,30 +221,52 @@ def exists(self) -> bool:
...
+            in the ORM class.
+
         Returns:
             ``True`` if a record was created, ``False`` if it was updated.
         """
         if self._deleted:
             raise RuntimeError(f"{self.id} was deleted")
+
         table = self.meta.table
-        fields = self.to_record(only_writable=True)["fields"]
```
gildastone, line 237: @BAPCon what do you think of below code? If basically keeps the same logic but filter the `Model.to_record` with the list of fields to save `fields_to_save`? Also, it adds some checks to make sure that the fields to save and declared in the Model ORM.

```
assert set(fields_to_save).issubset(self._attribute_desriptor_map().keys()), "Undeclared fields in Model."
fields = self.to_record(only_writable=True)["fields"]
fields = {k:v for fields.items() if k in fields_to_save}
```

The rest of this method is the same.

pyairtable/orm/model.py
```diff
@@ -322,6 +344,36 @@ def to_record(self, only_writable: bool = False) -> RecordDict:
         }
         ct = datetime_to_iso_str(self.created_time) if self.created_time else ""
         return {"id": self.id, "createdTime": ct, "fields": fields}
+    
+    def to_update_record(self, fields: Optional[List[str]] = None) -> UpdateRecordDict:
```
gildastone, line 348: @BAPCon same comment as above. Could be simplified by filtering the dictionary returned by `Model.to_record` with the list `fields: List[str]` of fields to update.

pyairtable/orm/model.py
```diff
@@ -373,7 +425,7 @@ def from_id(
             memoize: |kwarg_orm_memoize|
         """
         try:
-            instance = cast(SelfType, cls._memoized[record_id])
+            instance = cls._memoized[record_id]
```
gildastone, line 428: @BAPCon What's the reason of removing the casting?

tests/test_orm.py
```diff
@@ -71,6 +71,49 @@ def test_model_basics():
     assert record["fields"]["First Name"] == contact.first_name
 
 
+def test_model_save_specific_fields():
```
gildastone, line 74: @BAPCon Could be nice to add an additional test that fails the `Model.save` and `Model.batch_save` when an undeclared Model ORM field is passed in in the `fields`.

gildastone (commented):
@BAPCon Let me know your thoughts on my comments.

mesozoic:
I understand what you're going for here, but I would still prefer the approach of tracking changes to fields behind-the-scenes (so that callers don't need to keep track of which fields they're modifying). I can't think of any _other_ use cases where you'd want to avoid saving field values that were changed on the model, and adding this kwarg to the API will create debt that we'll need to support indefinitely.

If you really need this behavior today, you can do something like:
```python
update_fields = orm_obj.to_record(only_writable=True)["fields"]
update_fields = {k: v for (k, v) in update_fields.items() if k in SAVE_FIELDS}
orm_obj.meta.table.update(orm_obj.id, update_fields)
```

mesozoic:
I've posted #381 as an alternate proposal to address the issue. Open to questions or comments there.

PR #381, opened 2024-08-18
Track changes to ORM fields and only save modified fields (unless forced)

### LINKED ISSUE #378: Model.save() specific fields

mesozoic:
This branch adds change tracking to the ORM Model and Field classes, so that we can avoid sending field values to the API if they haven't been modified. This is the default behavior now; implementers can override this behavior by calling `obj.save(force=True)` on a model instance.

Fixes #378 

FYI @gildastone @bapcon

### DIFF
--- a/pyairtable/orm/changes.py
+++ b/pyairtable/orm/changes.py
@@ -1,0 +1,48 @@
+from typing import Callable, Iterable, List, SupportsIndex, Union
+
+from typing_extensions import TypeVar
+
+T = TypeVar("T")
+
+
+class ChangeNotifyingList(List[T]):
+    """
+    A list that calls a callback any time it is changed. This allows us to know
+    if any mutations happened to the lists returned from linked record fields.
+    """
+
+    def __init__(self, *args: Iterable[T], on_change: Callable[[], None]) -> None:
+        super().__init__(*args)
+        self._on_change = on_change
+
+    def __setitem__(self, index: SupportsIndex, value: T) -> None:  # type: ignore[override]
+        self._on_change()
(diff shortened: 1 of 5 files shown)


gildastone:
@mesozoic great work! Thanks for putting that nice solution 👏🏼 Can we close #379 ?

PR #386, opened 2024-08-25
Add `use_field_ids=` param to Api

### LINKED ISSUE #365: There should be a way to always set `return_fields_by_field_id` to `True` on all supported calls.
Is this already supported? If so, I have not been able to find it. If not, then it would be a welcome feature addition!

mesozoic:
This allows implementers to declare `use_field_ids=True` once, when creating the Api instance, and not have to pass it to each method call thereafter. FYI @jonathanlaniado 

Fixes #365


jonathanlaniado:
👏👏👏

PR #428, opened 2025-04-04
Release 3.1.0

mesozoic:


kkom:
@mesozoic – would there be any change to include a very small static types fix in this as well? 🙏 #429

PR #429, opened 2025-04-05
Fix `*args` static type hint in `FunctionCall` constructor

kkom:
This type hint is unnecessarily wrapped in `List[...]`.

The fact that `args` is a list (in fact, variable length tuple) is implicit.

This typing results in incorrectly typing `args` as `tuple[List[Any], ...]` – i.e. requires _each_ positional argument _itself_ be a `List[Any]`.

See https://peps.python.org/pep-0484/#arbitrary-argument-lists-and-default-argument-values 

### Test Plan

I don't see any static type checking in this project's CI at the moment: https://github.com/gtalarico/pyairtable/blob/132332d5905d5c5b7cc1eb7c01bc2bbffea3b4ff/.github/workflows/test_lint_deploy.yml

Therefore I just tested it manually. Specifically, I updated the `pyairtable` dependency in my personal project and Pyright stopped seeing a type error when calling instantiating `FunctionCall`.

Before:
 
Screenshot 2025-04-05 at 13 32 38

After:

Screenshot 2025-04-05 at 13 32 18

### DIFF
--- a/pyairtable/formulas.py
+++ b/pyairtable/formulas.py
@@ -524,7 +524,7 @@ class FunctionCall(Formula):
     for all formula functions known at time of publishing.
     """
 
-    def __init__(self, name: str, *args: List[Any]):
+    def __init__(self, name: str, *args: Any):
         self.name = name
         self.args = args
 


mesozoic:
Good catch, thanks! We do have static typing tests in `test_typing.py` but the liberal use of `Any` in the formulas module hid this from us. I'll add this to #428 since I'm about to ship it.

PR #433, opened 2025-05-12
Add Literal with date-time units for DATEADD and DATETIME_DIFF.

rafalkrupinski:


mesozoic:
Hi there! Thanks for the idea. It's an interesting thought to get more precise about the type annotations on these function signatures. There are probably other places where we could do something similar.

One challenge is that these are (currently) auto-generated by `make format` based on a [text file](https://github.com/gtalarico/pyairtable/blob/main/pyairtable/formulas.txt) which I extracted from Airtable's documentation. I'm hesitant to maintain this section by hand, because it'll be more time-consuming to update in the future when Airtable adds/changes their formula functions (or if we need to change how the library handles it).

rafalkrupinski:
Ah, you use cog for that... I didn't see that one coming ;)

Anyway, it shouldn't be too hard to extend the format with standard python type annotation syntax, accept `variable: type` and turn it to `variable: type | Formula` (`| None` for optional). WDYT

PR #451, opened 2026-05-01
enh: Add PyAirtable Caching Layer

taylorwilsdon:
This came up in a recent project that pulls a large amount of Airtable data and was taking 30-60 seconds to paint every time the page was called. To work around it, I just implemented a caching shim at the service level but figured we might as well bring the benefits directly into the library. I deliberately designed this as opt-in and as minimal as possible, so it should be a drop in enhancement that does not disrupt existing usage. 

## Summary

- Add opt-in TTL-based caching for `Table.all()` and `Table.first()` via `Api.enable_caching()`.
- Add `TableTTLConfig` for default/per-table TTLs and cache size limits, and expose cache diagnostics with `Api.cache_stats()`.
- Automatically invalidate cached table results after record or attachment mutations made through the same `Api` instance.
- Export and document the new caching configuration API.

## Tests

- Added unit and integration coverage for cache key normalization, TTL expiry, per-table overrides, deep-copy isolation, max-entry eviction, thread safety, disabled-by-default behavior, query separation, and mutation invalidation.

@mesozoic for viz

### DIFF
--- a/docs/source/api.rst
+++ b/docs/source/api.rst
@@ -18,6 +18,9 @@ API: pyairtable
 .. autoclass:: pyairtable.Table
     :members:
 
+.. autoclass:: pyairtable.TableTTLConfig
+    :members:
+
 .. autoclass:: pyairtable.Workspace
     :members:
 
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -5,6 +5,9 @@ Changelog
 3.3.0 (2025-11-05)
 ------------------------
 
+* Added opt-in TTL caching for ``Table.all`` and ``Table.first`` via
+  :meth:`Api.enable_caching <pyairtable.Api.enable_caching>` and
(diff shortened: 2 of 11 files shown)


mesozoic:
It's a valid problem but I'm not convinced this belongs in the clientlib itself. There are a lot of limitations to building this as a stateless feature around the request path - every thread will save its own duplicate cache, changes on the server can't invalidate, etc. Let's chat about options for a stateful proxy that sits in between the clientlib and the API which can be more adaptive.

taylorwilsdon:
Agree wholeheartedly that this should not be sold as a robust, adaptive caching layer. A stateful proxy likely is the better home if the goal is shared caching across workers/processes, coordinated invalidation, request coalescing, observability, or adaptive TTLs. Since the library already has endpoint_url support for debugging/caching proxies, probably the better approach here.

PR #461, opened 2026-07-25
Properly handle multi-codepoint emoji

Prill:
Emoji can consist of multiple Unicode codepoints. When these are used for reactions in AirTable, the API returns all codepoints, separated by hyphens. This handles that use case, which previously caused an error.

### DIFF
--- a/pyairtable/models/comment.py
+++ b/pyairtable/models/comment.py
@@ -123,7 +123,12 @@ class Reaction(AirtableModel):
         """
         The emoji character used for the reaction.
         """
-        return chr(int(self.emoji_info.unicode_character, 16))
+        return "".join(
+            [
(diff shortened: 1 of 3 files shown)


mesozoic:
It's strange that the API does this instead of just _returning the character_ (since JSON is already Unicode) but thanks for catching it!

Prill:
I guess they wanted something that was encoding-neutral? If you encode `"🧑‍🧑‍🧒‍🧒"` to JSON in Python, it escapes them as UTF-16 surrogate pairs, and I can see wanting to limit a JSON file to ASCII. Though I guess hyphen-separated code points is an encoding in its own right.

mesozoic:
This was just released as 3.4.1

